PR #118, opened 2021-09-16
WooComm Orders and Subscriptions migrator

iuravic:
Here's a new migrator which can migrate WooComm's Orders and Subscriptions (with all their belonging sub-objects) from one site to another.

The only purpose of this review is to let you know that this functionality exists here, in case you;d need it again. IMO no need for you to actually review this, because it's not being used in any live flow -- it's just here for when needed again.

The code is quite messy, because I wrote it in a couple of hours for the purposes of today's the Lisboa Mensagem launch.

Here's how it works:
- it expects DB tables from the source site to be alongside local WP DB tables, each with a different prefix
- takes list of Order IDs or Subscription IDs at the source site/DB, and exports&imports them into the WP DB
- works with pure SQL. Though it is inefficient, it made most sense to me at the time, because I was discovering the data relationships as I was working, and this was easiest to do with SQL. Plus it needs to pick through a different DB, which is also a reason why I used SQL.

### DIFF
--- a/newspack-custom-content-migrator.php
+++ b/newspack-custom-content-migrator.php
@@ -43,7 +43,7 @@ PluginSetup::register_migrators( array(
 	Migrator\General\ReusableBlocksMigrator::class,
 	Migrator\General\SportsPressMigrator::class,
 	Migrator\General\FeaturedImagesMigrator::class,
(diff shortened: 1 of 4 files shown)


kariae:
Thank you @iuravic I'll keep this one in mind :)

PR #124, opened 2021-10-14
feat: creating migrator for el libero posts.

eddiesshop:
El Libero has custom post_types for posts. These need to be migrated so that the post_type = post. This is so that all posts will be visible on their site. The custom post_type's have been migrated to be categories instead.

### DIFF
--- a/src/Migrator/PublisherSpecific/ElLiberoMigrator.php
+++ b/src/Migrator/PublisherSpecific/ElLiberoMigrator.php
@@ -114,7 +114,7 @@ class ElLiberoMigrator implements InterfaceMigrator {
 			WP_CLI::line( 'Linking ' . count( $post_ids ) . ' posts to new category.' );
 
 			foreach ( $post_ids as $post_id ) {
-				wp_set_post_categories( $post_id, $category->term_id );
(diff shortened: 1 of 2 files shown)


src/Migrator/PublisherSpecific/ElLiberoMigrator.php
```diff
@@ -0,0 +1,128 @@
...
+				$category = get_term( $category_id );
+			}
+
+			// Get all posts with post_type.
+			$post_ids = $wpdb->get_col( "SELECT ID FROM {$wpdb->posts} WHERE post_type = '{$post_type}'" );
+
+			WP_CLI::line( 'Linking ' . count( $post_ids ) . ' posts to new category.' );
+
+			foreach ( $post_ids as $post_id ) {
+				wp_set_post_categories( $post_id, $category->term_id );
```
claudiulodro, line 117: [You should pass in `true` as a third param here, otherwise it'll overrwrite any existing categories.](https://developer.wordpress.org/reference/functions/wp_set_post_categories/) If you pass in `true`, it'll append this category in addition to any other categories a post may already have.

claudiulodro (approved):
Thanks for making that update!

PR #128, opened 2021-12-22
feat: custom migrator for El Libero custom categories.

eddiesshop:
This PR is for a custom category migration script which establishes a hierarchy defined by the El Libero team for their categories from existing data.


iuravic:
Great work, @eddiesshop 👍 Seeing how you used SQL and were able to completely manage WP's DB structure and data makes me appreciate your skills and also assume you've gained a deep understanding of taxonomies and terms. I'm approving this without testing, because it looks good to me.

I would love us to talk about this some more in our LE Spring meeting, because it's a great example of a reusable functionality we should "productize" into a reusable LE tool (i.e. command or script) and improve our LE tech that way.

I recently also needed to remap a category structure for one of our Publishers, and this required managing hierarchical trees and their (changing) relational mapping, and all the data related to it. In my example, the source category tree and destination category tree had different depths, which made it additionally complex. Here's sharing [my code from this example](https://github.com/Automattic/newspack-custom-content-migrator/blob/master/src/Migrator/PublisherSpecific/GrehlakshmiMigrator.php#L485). I was able to use all the WP's built in functionalities (to get the categories, create them) and was able to avoid using any SQL, and I used recursive functions to scan and rebuild category trees.

For now I'd just like to point 2 thoughts which I'd love for us to talk about further in our LE call -- CC @kariae:
- how can we reuse and "productize" functionality like this best?
- how can we take advantage of all WP's built in functions and avoid more demanding and error sensitive (but practical) pure SQL approach?

iuravic (approved):

iuravic (approved):
Thanks for doing this, @eddiesshop 🙇 

Here's a recap of the workflow in this particular repo for future, CC @kariae. Publisher-specific migration code doesn't require a full review process, so the author can freely merge to master as they please. The author just tags their fellow LEs and shares the info that new code has landed, like this:
- author opens a PR,
- tags other LEs as reviewers (so that they get an email and can see that some new code landed),
- the author can immediately merge to master without waiting for review approval.

In case of other reusable code and commands (like general type migrators), we'll want to go through the regular review where the approval is required before merging to master.

PR #134, opened 2022-02-03
Add: umbria24 migrator

kariae:


kariae:
@iuravic here is the [function to generate Jetpack Slideshow Block from posts IDs](https://github.com/Automattic/newspack-custom-content-migrator/pull/134/files#diff-fd23d915387758f8715fa75368767262ffe59d9f7ca0c99229e6b30ca8081a92R483), and [an example](https://github.com/Automattic/newspack-custom-content-migrator/pull/134/files#diff-405e1767b8819a6db8f9bf4f443fe2963a71f3341c7517ee8d53a454b8dc6b26R291) of its usage.

PR #140, opened 2022-03-03
Feat(nextgen gallery migrator)

kariae:
This PR adds the possibility to migrate NextGen Gallery shortcode `[nggallery id=3392]` to either a Jetpack or Gutenberg Gallery block.

It lets also the previous NextGen Gallery Block migrator set the output to either Jetpack or Gutenberg Gallery blocks.

This is done by adding the following argument and flag to the WP CLI command:

- `gallery-block-type`: Gallery block type to be used as a replacement for the Next Geb Gallery shortcode. Accepted values are gutenberg-gallery, jetpack-gallery. Defaults to gutenberg-gallery. 
- `from-shortcode`: Flag to be set if the NGG shortcode is escaped in the post content.

080fe85bc5a77ce9c65083a244486da77466116a Fix some code styling based on WordPress Code Styling. Sorry to add this, as I have formatting automatically enabled. I can definitely revert this if needed, but I think we'll need at some point to fix the PHPCS config and use it 😅


src/Migrator/General/NextgenGalleryMigrator.php
```diff
@@ -78,11 +77,30 @@ public static function get_instance() {
...
+			'newspack-content-migrator nextgen-gallery-to-gutenberg-gallery-blocks',
+			array( $this, 'cmd_nextgen_galleries_to_gutenberg_gallery_blocks' ),
+			array(
 				'shortdesc' => 'Import NextGen images to Media Library, and converts NextGen Galleries throughout all Posts and Pages to Gutenberg Gallery blocks. Before running this command, make sure NextGen DB tables are present locally, and that the gallery folder is found at the configured location and contains all the NextGen images and has correct file permissions.',
-			]
+				'synopsis'  => array(
+					array(
+						'type'        => 'assoc',
+						'name'        => 'gallery-block-type',
+						'description' => 'Gallery block to be used as replacement for the Next Geb Gallery shortcode. Accepted values are: gutenberg-gallery, jetpack-gallery. Defaults to gutenberg-gallery.',
```
iuravic, line 89: A quick question, in your opinion should we default to gutenberg-gallery or jetpack-gallery instead? I'm thinking since all our sites are running with JP, perhaps we'd want to chose that, if it has more display abilities.

src/Migrator/General/NextgenGalleryMigrator.php
```diff
@@ -212,28 +230,33 @@ public function import_ngg_image_to_media_library( $image_row, $gallery_row ) {
...
 	 *
 	 * @return bool|\WP_Error
 	 */
-	public function convert_ngg_galleries_in_post_to_gutenberg_gallery( $post ) {
+	public function convert_ngg_galleries_in_post_to_block_gallery( $post, $from_shortcode, $gallery_block_type ) {
 		global $wpdb;
 
-		// Convert NGG Gallery blocks to pure Gutenberg Gallery blocks.
-		$post_content_updated = $this->convert_ngg_gallery_blocks_to_gutenberg_gallery_blocks( $post->ID, $post->post_content );
+		if ( $from_shortcode ) {
```
iuravic, line 242: Instead of this if block, could we perhaps just run one conversion after the other, like this:
```
...
$post_content_updated = $this->convert_ngg_gallery_shortcode_to_gallery_blocks( $post->ID, $post->post_content, $gallery_block_type );
$post_content_updated = $this->convert_ngg_gallery_blocks_to_gutenberg_gallery_blocks( $post->ID, $post->post_content, $gallery_block_type );
...
```

In that case, we could drop the `from-shortcode` CLI argument completely, and let the code replace both blocks and shortcodes if they're found.

iuravic (changes requested):
Beautifully done, Zak! 🙌
I just left a couple of questions before we finish up.

src/Migrator/General/NextgenGalleryMigrator.php
kariae: Yep, I agree with you. I think we can add the basic display abilities of the Jetpack Gallery Block via command flags because the current code uses the default block display features.

I'll switch the default block to jetpack-gallery in this PR, and reserve adding the display abilities to the next PR when we'll need them.

src/Migrator/General/NextgenGalleryMigrator.php
kariae: That was how I did it in the beginning, but while running tests it didn't work as expected. `convert_ngg_gallery_shortcode_to_gallery_blocks` returns a WP_Error in certain cases, one is [when there is no ngg shortcode](https://github.com/Automattic/newspack-custom-content-migrator/blob/master/src/Migrator/General/NextgenGalleryMigrator.php#L295) which is actually the case when NGGallery is used with a shortcode instead of a Block.

If we want to go this route, both conversion functions need to return the non-mutated post content in case of errors instead of a WP_Error, but in this case, we'll lose some logging information 🤔 .

kariae: I totally agree with you. Yes, I'll do some refactoring to remove the WP_Error returns and the CLI parameter in this PR, it will be useful to have the conversion in one command and more elegant!

kariae: @iuravic as discussed, I removed the unnecessary flag so now we'll convert blocks and shortcodes with one command.

I had a use case where I needed the images from the Gutenberg Gallery Block to show a lightbox on click, so I added a way to customize blocks. Instead of using flags to customize blocks, I used more options for the `gallery-block-type` argument (because some customization options would be unique per Block). This way, if we have to add more customization to the generated blocks or even new blocks in future it will be easy, what do you think?

https://github.com/Automattic/newspack-custom-content-migrator/pull/140/commits/de459294cb1cf305061d58a30a016466c8231363#diff-0013fe12ad1ca94b59b0671d549fe2faadcf7c3b6430573647ecc1d1533613a8R448-R459 
I had to remove the `<figcaption>` tag when the caption is empty because if we add an empty `<figcaption>` tag it breaks the block.

iuravic (approved):

src/Migrator/General/NextgenGalleryMigrator.php
```diff
@@ -78,11 +80,23 @@ public static function get_instance() {
...
+			'newspack-content-migrator nextgen-gallery-to-gutenberg-gallery-blocks',
+			array( $this, 'cmd_nextgen_galleries_to_gutenberg_gallery_blocks' ),
+			array(
 				'shortdesc' => 'Import NextGen images to Media Library, and converts NextGen Galleries throughout all Posts and Pages to Gutenberg Gallery blocks. Before running this command, make sure NextGen DB tables are present locally, and that the gallery folder is found at the configured location and contains all the NextGen images and has correct file permissions.',
-			]
+				'synopsis'  => array(
+					array(
+						'type'        => 'assoc',
+						'name'        => 'gallery-block-type',
+						'description' => 'Gallery block type to be used as a replacement for the Next Geb Gallery shortcode. Accepted values are gutenberg-gallery, jetpack-gallery. Defaults to gutenberg-gallery.',
```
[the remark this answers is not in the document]
kariae: Thank you, Ivan, for catching the typos 👍

PR #144, opened 2022-03-16
Feat: NinjaTables exporter

kariae:
This PR proposes a bulk exporter for the tables created with NinjaTables.

A first attempt to do this was by doing it manually (querying the tables ...) but I switched to use the same way we use with other migratory which is using the plugin already made classes/methods. So in order to work, the command needs the NinjaTables plugin to be installed.

As of now, there is only one command that does the bulk import if needed in the future we can add a parameter or a second command to export a specific table.

The command export to either CSV or JSON files (default to CSV). Below are a few examples to use the command:

```bash
# Export all tables to a given folder (/tmp) as CSV files.
wp newspack-content-migrator export-ninja-tables --output-dir=/tmp
# Export all tables to a given folder (/tmp) as JSON files.
wp newspack-content-migrator export-ninja-tables --output-dir=/tmp --output-type=json
```

@iuravic @eddiesshop  I'll need a code review since it introduces a migration logic class too (No rush :)).

### DIFF
--- a/newspack-custom-content-migrator.php
+++ b/newspack-custom-content-migrator.php
@@ -47,6 +47,7 @@ PluginSetup::register_migrators( array(
 		Migrator\General\ContentDiffMigrator::class,
 		Migrator\General\WooCommOrdersAndSubscriptionsMigrator::class,
 		Migrator\General\NextgenGalleryMigrator::class,
+		Migrator\General\NinjaTablesMigrator::class,
 
 		// Publisher specific:
 		Migrator\PublisherSpecific\GadisMigrator::class,
--- a/src/MigrationLogic/NinjaTables.php
+++ b/src/MigrationLogic/NinjaTables.php
@@ -70,7 +70,7 @@ class NinjaTables {
 				}
 				array_push( $export_data, $temp );
 			}
-			$this->export_as_csv( $export_data, $file_path );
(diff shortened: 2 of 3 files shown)


iuravic (changes requested):
Great work, @kariae 🙇 I love the approach with utilizing NinjaTables' Plugin's code to do the work for us 👍

The logic class is perfect. I'd just suggest we added one small error message to the command class too.

[This comment line here](https://github.com/Automattic/newspack-custom-content-migrator/pull/144/files#diff-1cb7ccbfd4dd329ca99de47b6b4e28710260635f592d6f0405d3c2a6aab2192bR32) says everything. Could we add a quick check at the beginning of the [`cmd_export_ninja_tables` function](https://github.com/Automattic/newspack-custom-content-migrator/pull/144/files#diff-a5b82b44e4a6ede89c0c882b81b1c0a207edb4196d5a99580667c84aacae9f97R81) something like:
```
if ( is_null( $this->ninja_tables_logic->ninja_tables_admin ) ) {
  WP_Error( 'Ninja Tables plugin is a dependency, and will have to be installed before this command can be used.' );
}
```

It's a tiny thing I think would make this entirely perfect :)

iuravic (approved):

kariae:
Thank you @iuravic that was a quick one 🙇

PR #146, opened 2022-03-22
Add/split guest author

eddiesshop:
Takes a guest author string as the first argument. All posts that guest author has been associated with will then get reassigned to whatever new guest authors have been supplied to be created.

### DIFF
--- a/src/Migrator/General/CoAuthorPlusMigrator.php
+++ b/src/Migrator/General/CoAuthorPlusMigrator.php
@@ -220,19 +220,19 @@ class CoAuthorPlusMigrator implements InterfaceMigrator {
 			'newspack-content-migrator co-authors-split-to-multiple-coauthors',
 			[ $this, 'cmd_split_guest_author_to_multiple_authors' ],
 			[
-				'shortdesc' => 'Used to take a guest author and split to multiple different guest authors the user specifies',
+				'shortdesc' => 'Reassigns all posts currently associated to a specific Guest Author to different multiple (or single) Guest Author(s)',
 				'synopsis' => [
 					[
 						'type' => 'positional',
 						'name' => 'guest-author',
-						'description' => 'Target Guest Author (string)',
+						'description' => 'Target Guest Author (Guest Author Display Name e.g. Edward Carrasco with Newspack, and Ingrid Miller with HR)',
 						'optional' => false,
 						'repeating' => false,
 					],
 					[
 						'type' => 'assoc',
 						'name' => 'new-guest-author-names',
-						'description' => "The new guest author's to create (delimited values accepted). The original will be disassociated.",
+						'description' => "The new guest author's to create (e.g. Edward Carrasco,Ingrid Miller [delimited by comma]). The original GA will be disassociated.",
 						'optional' => false,
 						'repeating' => true,
 					],
@@ -696,7 +696,7 @@ class CoAuthorPlusMigrator implements InterfaceMigrator {
 		global $wpdb;
 		global $coauthors_plus;
 
-		$guest_author_posts = "SELECT tr.* FROM $wpdb->term_relationships tr 
+		$guest_author_posts_sql = "SELECT tr.* FROM $wpdb->term_relationships tr 
 			INNER JOIN $wpdb->posts p ON tr.object_id = p.ID
 			WHERE tr.term_taxonomy_id IN (
 	            SELECT wtt.term_taxonomy_id FROM $wpdb->term_taxonomy wtt
@@ -705,7 +705,7 @@ class CoAuthorPlusMigrator implements InterfaceMigrator {
 	                SELECT post_id FROM $wpdb->postmeta WHERE meta_key = 'cap-display_name' AND meta_value = '$old_guest_author'
 	            )
 			) AND p.post_type = 'post'";
-		$results = $wpdb->get_results( $guest_author_posts );
+		$results = $wpdb->get_results( $wpdb->prepare( $guest_author_posts_sql ) );
 
 		if ( empty( $results ) ) {
 			WP_CLI::success('Target guest author not found.');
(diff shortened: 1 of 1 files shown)


src/Migrator/General/CoAuthorPlusMigrator.php
```diff
@@ -215,6 +216,37 @@ public function register_commands() {
...
+						'type' => 'positional',
+						'name' => 'guest-author',
+						'description' => 'Target Guest Author (string)',
+						'optional' => false,
+						'repeating' => false,
+					],
+					[
+						'type' => 'assoc',
+						'name' => 'new-guest-author-names',
+						'description' => "The new guest author's to create (delimited values accepted). The original will be disassociated.",
```
iuravic, line 235: Ah, so the command will create the new GAs, too. Very nice! I see it in the code now after reading this, but it helped a lot that it's stated in this description. Thanks!

Let's please add an example to the end of this "description" sentence, too. I'd suggest adding both an example for multiple GAs and for a single GA, just to demonstrate the power of this command a bit more.

src/Migrator/General/CoAuthorPlusMigrator.php
```diff
@@ -215,6 +216,37 @@ public function register_commands() {
...
+						'optional' => false,
+						'repeating' => true,
+					],
+					[
+						'type' => 'assoc',
+						'name' => 'delimiter',
+						'description' => 'Delimiter to use to split new-guest-authors values.',
```
iuravic, line 242: I wonder would it have made sense the delimiter was standardized, like if we used a comma? Perhaps that way:
- we wouldn't need this parameter,
- we could say that the `new-guest-author-names` is CSV

src/Migrator/General/CoAuthorPlusMigrator.php
```diff
@@ -650,6 +682,93 @@ public function cmd_cap_fix_non_unique_guest_slugs() {
...
+	 */
+	public function cmd_split_guest_author_to_multiple_authors( $args, $assoc_args ) {
+		$old_guest_author = $args[0];
+		$delimiter = $assoc_args['delimiter'];
+		$new_guest_author_names = explode( $delimiter, $assoc_args['new-guest-author-names'] );
+
+		global $wpdb;
+		global $coauthors_plus;
+
+		$guest_author_posts = "SELECT tr.* FROM $wpdb->term_relationships tr
```
iuravic, line 699: Let's use `$wpdb->prepare` to prepare the SQL parameter `$old_guest_author`, please, to prevent SQL injection.

It's a very far fetched case that this could be misused, since we use it for our local work internally, but `$wpdb->prepare` is so handy and we can rest easy :)

src/Migrator/General/CoAuthorPlusMigrator.php
```diff
@@ -650,6 +682,93 @@ public function cmd_cap_fix_non_unique_guest_slugs() {
...
+							'user_login'   => $sanitized_guest_author_name,
+							'display_name' => sanitize_text_field( $guest_author_name ),
+						]
+					);
+				} else {
+					$author_id = $user->ID;
+				}
+			}
+
+			$guest_author = ( new CoAuthors_Guest_Authors() )->get_guest_author_by( 'id', $author_id );
```
iuravic, line 740: This is perfectly fine, just sharing that alternatively we could have used `$this->coauthorsplus_logic->coauthors_guest_authors` since that's already an instantiated [public property in the `$this->coauthorsplus_logic`](https://github.com/Automattic/newspack-custom-content-migrator/blob/master/src/MigrationLogic/CoAuthorPlus.php#L41). But this works well just the same!

iuravic (changes requested):
Looks great, and could come in really handy! Thanks for making this into a reusable command!

I left just a couple of very small notes here.

src/Migrator/General/CoAuthorPlusMigrator.php
```diff
@@ -215,6 +216,37 @@ public function register_commands() {
 		WP_CLI::add_command( 'newspack-content-migrator co-authors-fix-non-unique-guest-slugs', array( $this, 'cmd_cap_fix_non_unique_guest_slugs'), [
 		    'shortdesc' => "Make unique any Guest Author Slug which matches a User's slug."
         ] );
+		WP_CLI::add_command(
+			'newspack-content-migrator co-authors-split-to-multiple-coauthors',
+			[ $this, 'cmd_split_guest_author_to_multiple_authors' ],
+			[
+				'shortdesc' => 'Used to take a guest author and split to multiple different guest authors the user specifies',
```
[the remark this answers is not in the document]
eddiesshop: Great, thanks for suggestion!

src/Migrator/General/CoAuthorPlusMigrator.php
eddiesshop: I'd like to leave it as-is, I really meant for this to be a one-off command that we can run from the CLI. If we need to add some mass update functionality, we can update this to support a CSV.

src/Migrator/General/CoAuthorPlusMigrator.php
eddiesshop: Updated!

src/Migrator/General/CoAuthorPlusMigrator.php
eddiesshop: Thanks! Updated.

src/Migrator/General/CoAuthorPlusMigrator.php
```diff
@@ -215,6 +216,37 @@ public function register_commands() {
 		WP_CLI::add_command( 'newspack-content-migrator co-authors-fix-non-unique-guest-slugs', array( $this, 'cmd_cap_fix_non_unique_guest_slugs'), [
 		    'shortdesc' => "Make unique any Guest Author Slug which matches a User's slug."
         ] );
+		WP_CLI::add_command(
+			'newspack-content-migrator co-authors-split-to-multiple-coauthors',
+			[ $this, 'cmd_split_guest_author_to_multiple_authors' ],
+			[
+				'shortdesc' => 'Reassigns all posts currently associated to a specific Guest Author to different multiple (or single) Guest Author(s)',
```
iuravic, line 223: Let's please update this description to the following:
```
Reassigns all posts currently associated to a specific Guest Author to different multiple (or single) Guest Author(s). Example use: `wp newspack-content-migrator co-authors-split-to-multiple-coauthors "Chris Christofersson of NewsSource, with Adam Black of 100 Days in Appalachia" --new-guest-author-names="Chris Christofersson; Adam Black of 100 Days in Appalachia" --delimiter="; "
```

iuravic (changes requested):

PR #154, opened 2022-04-29
Add/reno postgress archive importer

iuravic:
Custom content migrator for Reno, Django Python's Postgres DB to WP.


src/Migrator/PublisherSpecific/RenoMigrator.php
```diff
@@ -0,0 +1,395 @@
...
+				}
+			}
+		} else {
+			$msg = sprintf( 'ERR featImg not found, $post_id=%d, $img_path=%s', $post_id, $featimg_full_path );
+			$this->log( $log, $msg );
+			\WP_CLI::line( $msg );
+		}
+	}
+
+	public function get_or_create_guest_author( $post_id, $log, $images_path, $author_id, $author_name, $author_email, $author_bio ) {
```
kariae, line 322: I think that this is a useful function that we can refactor to the CoAuthorPlus logic class, as it can be used for other publishers.

kariae (approved):
LGTM.

src/Migrator/PublisherSpecific/RenoMigrator.php
iuravic: I'm just examining this option. At first it seemed perfectly logical, and I immediately created a new branch to refactor this.

But then I noticed that my method here is basically just a wrapper for [create_guest_author](https://github.com/Automattic/newspack-custom-content-migrator/blob/master/src/MigrationLogic/CoAuthorPlus.php#L97) which is already reusable, and just does two things:
- imports the avatar image to Attachment to get att. ID
- does some custom logging for Reno purposes
- does several unnecessary extra validation for $args array which goes in CoAuthorPlus::create_guest_author

So on second look, perhaps we don't need to share this one.

Let me know if I've missed something! 🙇

kariae: You're making a point here. I think all's good, thank you.

PR #158, opened 2022-05-20
Feat/add new orleans tribune migrator

kariae:
A migrator of WPPA albums to Jetpack Slideshow Block, this is set for one publisher, but can ve easily refactored later to be used for other publishers.

### DIFF
--- a/src/Migrator/General/PhotoAlbumsMigrator.php
+++ b/src/Migrator/General/PhotoAlbumsMigrator.php
@@ -1,0 +1,225 @@
+<?php
+
+namespace NewspackCustomContentMigrator\Migrator\General;
+
+use \NewspackCustomContentMigrator\Migrator\InterfaceMigrator;
+use \NewspackCustomContentMigrator\MigrationLogic\Posts as PostsLogic;
+use \NewspackContentConverter\ContentPatcher\ElementManipulators\SquareBracketsElementManipulator;
+use \NewspackPostImageDownloader\Downloader;
+use \WP_CLI;
+
+/**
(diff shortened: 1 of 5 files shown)


iuravic (approved):
It would be great if we had this in a reusable form, @kariae :) Do you think it would be much effort to extract it into a reusable migrator?

kariae:
@iuravic I add the possibility to handle many albums' shortcodes within the same post content and set this as general migratory. The tests on the publisher content were successful.

src/Migrator/General/PhotoAlbumsMigrator.php
```diff
@@ -0,0 +1,225 @@
...
+use \NewspackCustomContentMigrator\Migrator\InterfaceMigrator;
+use \NewspackCustomContentMigrator\MigrationLogic\Posts as PostsLogic;
+use \NewspackContentConverter\ContentPatcher\ElementManipulators\SquareBracketsElementManipulator;
+use \NewspackPostImageDownloader\Downloader;
+use \WP_CLI;
+
+/**
+ * Custom migration scripts for Photo Albums Plugins.
+ */
+class PhotoAlbumsMigrator implements InterfaceMigrator {
```
iuravic, line 14: Looks amazing! 💯 💯 💯 ! 🥳 !
Just one thing, could we please rename `PhotoAlbumsMigrator` to something just a tiny bit more descriptive, like `PhotoAlbumProGalleryMigrator`?

iuravic (changes requested):

iuravic (approved):

PR #159, opened 2022-05-20
PMP 2 WooComm migrator

iuravic:


iuravic:
Merged in new https://github.com/Automattic/newspack-custom-content-migrator/pull/533

PR #161, opened 2022-05-27
feat(SearchLight NM): remove shortcodes from post content and migrate subtitles to meta

kariae:


iuravic (approved):
LGTM 👍 

If another review after solving the newspack-custom-content-migrator.php conflict is needed, feel free to merge it even without another approval 🙇

PR #162, opened 2022-05-31
feat: created a function to seek missing images in <img> tags.

eddiesshop:
This function will look at a site's wp_posts table for any posts with  tags in their post_content field. Out of the posts that it does find, it will load the post_content as HTML and cycle through any  tags. If it cannot find the indicated file in the src attribute, it will attempt to either look for it locally in the wp-content/uploads folder, or attempt to download it. If the image is downloaded, it will attach it to the post, and update the HTML content. The HTML content is also updated if the file is found locally as well.

### DIFF
--- a/src/Migrator/General/FixMissingMedia.php
+++ b/src/Migrator/General/FixMissingMedia.php
@@ -70,6 +70,12 @@ class FixMissingMedia implements InterfaceMigrator {
 				],
 			],
 		] );
+		WP_CLI::add_command( 'newspack-content-migrator fix-incomplete-sources',
+			[ $this, 'cmd_fix_incomplete_sources' ],
+			[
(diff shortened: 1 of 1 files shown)


iuravic (changes requested):
Thanks for the awesome PR, @eddiesshop  -- the idea, initiative, hands on approach, practicality, everything! :high-five :) ☀️

For purposes of code reusability and best productization for this functionality, let's examine what's the best place for us to host this code. We do already have these two plugins which do a similar and related function:
- https://github.com/Automattic/newspack-post-image-downloader
- https://github.com/Automattic/fix-missing-media

Let's please take a moment to discuss our path with this in Slack -- avoiding duplication, and best way to unify our existing functionality, and then keep extending it with new features, of course -- and come back to this PR when we decide as a Team.

For now I suggest we mark this with "Request changes" review status, to signify that we're considering all our options ATM off board (in Slack).

PR #163, opened 2022-05-31
newsroom.co.nz custom migrator

iuravic:
Reads postmeta saved by the Scraper Migrator Plugin, and turns it into the actual data that it should be.

### DIFF
--- a/src/Migrator/PublisherSpecific/NewsroomCoNzMigrator.php
+++ b/src/Migrator/PublisherSpecific/NewsroomCoNzMigrator.php
@@ -88,37 +88,37 @@ class NewsroomCoNzMigrator implements InterfaceMigrator {
 			$author_avatar_src = get_post_meta( $post->ID, 'newspackscraper_authoravatarsrc', true );
 			$feat_img_caption = get_post_meta( $post->ID, 'newspackscraper_featimgcaption', true );
 
-			// // Create GA with bio and avatar, link w/ WP User Author, and assign to Post.
-			// if ( $author_bio || $author_avatar_src ) {
-			// 	$author_name = get_the_author_meta( 'display_name', $post->post_author );
-			// 	$author_wp_user = get_user_by( 'id', $post->post_author );
-			// 	$ga_existing = $this->cap_logic->get_guest_author_by_user_login( $author_wp_user->user_nicename );
-			// 	if ( $ga_existing ) {
-			// 		$ga_id = $ga_existing->ID;
-			// 	} else {
-			// 		$avatar_att_id = $this->attachment_logic->import_external_file( $author_avatar_src );
-			// 		$ga_id = $this->cap_logic->create_guest_author( [
-			// 			'display_name' => $author_name,
-			// 			'description' => $author_bio,
-			// 			'avatar' => $avatar_att_id,
-			// 		] );
-			// 	}
-			// 	$this->cap_logic->link_guest_author_to_wp_user( $ga_id, $author_wp_user );
-			// 	$this->cap_logic->assign_guest_authors_to_post( [ $ga_id ], $post->ID );
-			// } else {
-			// 	// TODO debug
-			// 	$d=1;
-			// }
-			//
-			// // Feat img caption.
-			// if ( $feat_img_caption ) {
-			// 	$post_thumbnail_id = get_post_thumbnail_id( $post );
-			// 	$wpdb->update(
-			// 		$wpdb->posts,
-			// 		[ 'post_excerpt', $feat_img_caption ],
-			// 		[ 'ID' => $post_thumbnail_id ]
-			// 	);
-			// }
+			// Create GA with bio and avatar, link w/ WP User Author, and assign to Post.
+			if ( $author_bio || $author_avatar_src ) {
+				$author_name = get_the_author_meta( 'display_name', $post->post_author );
+				$author_wp_user = get_user_by( 'id', $post->post_author );
+				$ga_existing = $this->cap_logic->get_guest_author_by_user_login( $author_wp_user->user_nicename );
+				if ( $ga_existing ) {
+					$ga_id = $ga_existing->ID;
+				} else {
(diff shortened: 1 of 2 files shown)


src/Migrator/PublisherSpecific/NewsroomCoNzMigrator.php
```diff
@@ -0,0 +1,184 @@
...
+				$path_exploded = explode( '/', $path );
+				$slug = $path_exploded[1];
+				$wpdb->update(
+					$wpdb->posts,
+					[ 'post_name' => $slug ],
+					[ 'ID' => $post->ID ]
+				);
+			} else {
+				// TODO debug
+				$d=1;
```
kariae, line 169: I guess we can add a log here instead. That's said, I'm not sure about the use case here, but we can create the co-authors without their bio and avatars, as long as we do have the at least the author name, does it make sense to just remove the condition?

src/Migrator/PublisherSpecific/NewsroomCoNzMigrator.php
```diff
@@ -0,0 +1,184 @@
...
+				'newsroom.co.nz/environment' => 'Environment',
+				'newsroom.co.nz/business' => 'Business',
+				'newsroom.co.nz/comment' => 'Comment',
+				'newsroom.co.nz/technology' => 'Technology',
+				'newsroom.co.nz/new-auckland' => 'Auckland',
+				'newsroom.co.nz/health--science' => 'Health & Science',
+				'newsroom.co.nz/podcasts' => 'Podcasts',
+			];
+			foreach ( $cat_urls_to_cat_names as $cat_url => $cat_name ) {
+				if ( true === str_contains( $original_url, $cat_url ) ) {
```
kariae, line 140: str_contains is a PHP8 feature, while the Atomic servers are still on 7.4, so this will break once executed there. I had to create a polyfill for this function (and others) since their use for other migratory, so I wonder if we can start the talk about migration to PHP 8?

src/Migrator/PublisherSpecific/NewsroomCoNzMigrator.php
```diff
@@ -0,0 +1,184 @@
...
+
+					$parent_category_id = wp_create_category( $parent_category_name, 0 );
+					$child_category_id = wp_create_category( $child_category_name, $parent_category_id );
+
+					wp_set_post_categories( $post->ID, [ $parent_category_id, $child_category_id ], false );
+				}
+			}
+
+			// Update permalink.
+			$original_url_parsed = parse_url( $original_url );
```
kariae, line 155: I believe the permalink migration is working, can you, please, add an example(as a comment) of the possible $original_url value, as I tried with a live URL of a post from the live site, and I had an error, I probably get the wrong URL format.

src/Migrator/PublisherSpecific/NewsroomCoNzMigrator.php
kariae, line 169: Same. we can add a log here :)

src/Migrator/PublisherSpecific/NewsroomCoNzMigrator.php
```diff
@@ -0,0 +1,184 @@
...
+			}
+
+			// Delete the scraper postmeta.
+			// delete_post_meta( $post->ID, 'newspackscraper_authorbio' );
+			// delete_post_meta( $post->ID, 'newspackscraper_authoravatarsrc' );
+			// delete_post_meta( $post->ID, 'newspackscraper_featimgcaption' );
+			// delete_post_meta( $post->ID, 'newspackscraper_url' );
+		}
+
+		// Clean up empty categories.
```
kariae, line 179: Is this a TODO?

kariae (changes requested):

src/Migrator/PublisherSpecific/NewsroomCoNzMigrator.php
iuravic: @kariae you're absolutely right, I fully agree with your remakrs if this was production code 👍  No need for these kinds of loops, they're messy.

The reason I left these helper snippets it is that I'll be the only one running this migrator (Publisher's LE), and it's a two-time job, so I've felt free to leave these helper snippets which make my workflow more efficient. Specifically, here I use step debugging, and these else-s are here for that purpose.

So, here's my question -- for PublisherSpecific Migrators, are we to review just the effect our code has on the Atomic infrastructure (the MySQL restrictions) and make sure it doesn't overload an Atomic machine, or should we start reviewing PublisherSpecific code as actual production code and do full reviews?

I'm absolutely OK with us doing either of these, and think that we should decide as a Team. I personally would vote that we keep this review arrangement:
- general and reusable code -- we review fully with two goals in mind: to make sure impact on the platform is OK, and that the code is completely in line with our best practices and standards,
- PublisherSpecific -- we just review impact on Platform, and let the LE merge at their convenience.

src/Migrator/PublisherSpecific/NewsroomCoNzMigrator.php
iuravic: Oh, so you mean the permalink migration is built into the Scraper? 😮

src/Migrator/PublisherSpecific/NewsroomCoNzMigrator.php
iuravic: Yes, exactly. For now I just imported the first 50 demo posts for the Publisher to review. They're not yet onboarded to Newspack, so I wanted to leave all the cats for dev purposes. If they agree they will join Newspack, I'll do an actual scrape and fill in this bit.

src/Migrator/PublisherSpecific/NewsroomCoNzMigrator.php
kariae: TIL, thank you for sharing the info, Ivan.

PR #168, opened 2022-06-07
Feat/clean content from unused shortcodes

kariae:
We had a couple of requests to clean content from non-used shortcodes, this PR add a command to the PostsMigrator to remove a list of shortcodes from all the posts/pages or a given list of posts/pages.

My text editor keeps on auto-formatting the code because we do have a phpcs configuration file which is not configured as it seems. 79eab3aada1cc20c1e2d39c9562075a44a681136 introduce a few changes on the formatting side of things.

913889f4e22b16c4bf3fe81a0e990358f40310b1 is the commit regarding the shorcodes removal.


iuravic (changes requested):
Looks great! Let's please just create a new migrator, something like ContenFixerMigrator, or something better.

iuravic (approved):

PR #175, opened 2022-06-21
Initial XMLMigrator class with a demo parsing method

iuravic:
Created the initial version of General type XMLMigrator.

Shared just one demo method which shows how we can use PHP to parse XML, traverse nodes and extract data.


kariae (approved):
Thank you for this, Ivan. This should be a nice base for us to build on.

PR #176, opened 2022-06-23
Feat/woo subscriptions payment method switcher

kariae:
This PR adds a new command to migrate WooCommerce Subscriptions payment methods from a given payment method to another.

It currently supports PayPal and Bank Transfers (bacs), we can simply add more payment methods when needed, we just need to make sure we take into consideration moving the payment methods meta (which is not supported today), for example, how to deal with Stripe Customer IDs if we're migrating to PayPal, but I don't think we'll have to deal with such cases.

The command offer also the ability to set the subscription expiry date.


kariae:
Closing this PR, as the code didn't work for love stores (it works only when the stores are on staging mode), it's seems related to the PayPal payment method. I ended up doing this manually for my case.

PR #177, opened 2022-06-25
Creates a skeleton version of the JP Tiled Gallery block

iuravic:
Added a method which creates a (skeleton version of the) JP Tiled Gallery block from attachment IDs.

Since the JP tiled gallery is created very dynamically by Gutenberg, by creating a dynamic layout of columns and rows based on images' orientation and sizes, this is very difficult to reproduce in the backend.

As a first step towards creating JP Tiled Gallery blocks, this method creates a "skeleton" block code, which can be saved to post_content, and can then be repaired/"recovered" by Gutenberg.

Before, the code this "skeleton" method produces:
image

After, the correct syntax repaired by Gutenberg:
image

### DIFF
--- a/src/MigrationLogic/Posts.php
+++ b/src/MigrationLogic/Posts.php
@@ -9,6 +9,10 @@ class Posts {
 	/**
 	 * Gets IDs of all the Pages.
 	 *
+	 * @param string $post_type   Post type.
+	 * @param array  $post_status Post status.
+	 * @param bool   $nopaging    nopaging $arg param for \WP_Query( $arg ).
+	 *
 	 * @return array Pages IDs.
 	 */
 	public function get_all_posts_ids( $post_type = 'post', $post_status = [ 'publish', 'future', 'draft', 'pending', 'private', 'inherit' ], $nopaging = true ) {
@@ -124,15 +128,15 @@ SQL;
 	 *      }
 	 * ```
 	 *
+	 * @param string $taxonomy   Taxonomy.
+	 * @param int    $term_id    term_id.
 	 * @param array  $post_types Post types.
-	 * @param string $taxonomy Taxonomy.
-	 * @param int    $term_id term_id.
 	 *
 	 * @return \WP_Post[]
 	 */
 	public function get_post_objects_with_taxonomy_and_term( $taxonomy, $term_id, $post_types = array( 'post', 'page' ) ) {
 		return get_posts(
-            [
+			[
 				'posts_per_page' => -1,
 				// Target all post_types.
 				'post_type'      => $post_types,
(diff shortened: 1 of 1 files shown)


kariae (changes requested):
Hi @iuravic 👋  I had a hard time to find the migrator I created for Tiles Galleries because I didn't push it yet, sorry for the late review 😰.

I had the same issue where the block generated was broken, and used the same workflow before finding it to be a heavy lift to do it for a site with a couple of hundreds of galleries.

The issue was in the tile size that I couldn't generate, and I see you're not adding it too here. To fix the issue, I did two things, generate the tile size depending on its index (the size will be fixed automatically to create a nice tiled gallery). The second thing, I used the `serialize_blocks` to generate the HTML code to insert into the post content instead of manually writing it (not sure if this helped with this case, but in general I do rely on this way to generate the code).

Here is a snippet of a working function, please feel free to take the tile size part of it, and see if it can help ease the proposed workflow.

```php
$gallery_content = '
<div class="wp-block-jetpack-tiled-gallery aligncenter is-style-rectangular">
<div class="tiled-gallery__gallery">
	<div class="tiled-gallery__row">
';

$gallery_content .= join(
' ',
array_map(
	function( $index, $attachment_id ) {
		$attachment_url = wp_get_attachment_url( $attachment_id );
		$tile_size      = $this->get_tile_image_size_by_index( $index );
		return '
		<div class="tiled-gallery__col" style="flex-basis: ' . $tile_size . '%">
			<figure class="tiled-gallery__item">
				<img
					alt="' . get_the_title( $attachment_id ) . '"
					data-id="' . $attachment_id . '"
					data-link="' . $attachment_url . '"
					data-url="' . $attachment_url . '"
					src="' . $attachment_url . '"
					data-amp-layout="responsive"
				/>
			</figure>
		</div>';
	},
	array_keys( $gallery_meta ),
	$gallery_meta
)
);

$gallery_content .= '
		</div>
	</div>
</div>
';

$gallery_block = array(
'blockName'    => 'jetpack/tiled-gallery',
'attrs'        => array(
	'columnWidths' =>
		array(), // It will be filled after fixing the block manually from the backend.
	'ids'          => $gallery_meta,
),
'innerBlocks'  => array(),
'innerHTML'    => $gallery_content,
'innerContent' => array( $gallery_content ),
);

/**
* Generate tile size based on its index.
*
* @param int $index Tile size.
* @return string
*/
private function get_tile_image_size_by_index( $index ) {
$tile_image_sizes = array( '66.79014', '33.20986', '33.33333', '33.33333', '33.33333', '50.00000', '50.00000' );
return $tile_image_sizes[ $index % count( $tile_image_sizes ) ];
}
```

iuravic:
Closing this PR in favor of @kariae 's [better code above](https://github.com/Automattic/newspack-custom-content-migrator/pull/177#pullrequestreview-1022763779) -- we'll open a new PR with that instead.

PR #178, opened 2022-06-27
feat: created an automated process to remove dupe data from meta table.

eddiesshop:
This function was created to help automate the reduction of a large wp_postmeta table which could adversely affect performance. It does this by looking at every Post ID in wp_postmeta and seeing if there's a difference between the total row count and the unique meta_keys. If there is a difference, it then creates a dynamic key by concatenating meta_key and meta_value, and deleting any rows where the dynamic key has more than 1 row existing in the table.

### DIFF
--- a/src/Migrator/General/ContentFixerMigrator.php
+++ b/src/Migrator/General/ContentFixerMigrator.php
@@ -86,6 +86,15 @@ class ContentFixerMigrator implements InterfaceMigrator {
 				),
 			)
 		);
+
+		WP_CLI::add_command(
+			'newspack-content-migrator remove-duplicate-post-meta-rows',
+			array( $this, 'remove_duplicate_post_meta_rows' ),
+			array(
+				'shortdesc' => 'This will remove any duplicate data that may exist in wp_postmeta table.',
+				'synopsis' => array(),
+			)
+		);
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


src/Migrator/General/ContentFixerMigrator.php
```diff
@@ -224,6 +233,69 @@ function( $inner_content ) use ( $shortcodes ) {
...
+	 * then creates a dynamic key by concatenating meta_key and meta_value, and deleting any
+	 * rows where the dynamic key has more than 1 row existing in the table.
+	 *
+	 * @param $args
+	 * @param $assoc_args
+	 */
+	public function remove_duplicate_post_meta_rows( $args, $assoc_args ) {
+		global $wpdb;
+
+		$results         = $wpdb->get_results( "SELECT p.ID as post_id, count(pm.meta_id) as counter, count(distinct pm.meta_key) as unique_keys FROM $wpdb->posts p INNER JOIN $wpdb->postmeta pm ON p.ID = pm.post_id GROUP BY p.ID HAVING counter <> unique_keys ORDER BY counter DESC" );
```
iuravic, line 249: Hi @eddiesshop , thanks so much for putting this together and sharing 💯 ! Looking at this code, it seems to me that it's already doing a great job and doing the job correctly. I just wonder if we can perhaps do a second pass and clean it up a bit, by decreasing the number of queries and maybe making it a bit more performant and readable for future changes.

I basically have just one suggestion, to combine the first two queries into one query. @kariae, could you please also check this and share your thoughts?

So the part which [cleans the dupes, lines 249-278](https://github.com/Automattic/newspack-custom-content-migrator/pull/178/files#diff-89371c8d4b14fb806dcd4d00308135320cf686634ebfe42ece522c55cd02f9a1R249-R278) could maybe also be done with a simpler single query, something like:
```
select wpm.post_id, wpm.meta_id
from wp_postmeta wpm
where wpm.post_id in (
	select wpm2.post_id
	from wp_postmeta wpm2
	group by wpm2.post_id, wpm2.meta_key
	having count(*) > 1
)
order by wpm.post_id, meta_id;
```

where the inner query gets post_ids which have duplicate meta_keys, and the outer query gets which those specific meta_ids per post are.

Let's say that our wp_postmeta contains these rows:
| meta_id | post_id | meta_key | meta_value |
| ------------- | ------------- |------------- |------------- |
| 1 | 2 | _wp_page_template | default |
| 2 | 3 | _wp_page_template | default |
| 3 | 2 | _wp_page_template | new |
| 4 | 5 | aaa | 1 |
| 5 | 5 | aaa | 2 |
| 6 | 5 | aaa | 3 |

This would output all dupes for post_id 2 and 5 like this:
| post_id | meta_id |
|------------- |------------- |
| 2 | 1 |
| 2 | 3 |
| 5 | 4 |
| 5 | 5 |
| 5 | 6 |

So the next step would be to loop through these meta_ids by post_ids, and leave just the latest/newest meta_ids while deleting previous/older meta_ids.

Or it could even be decorated with your GROUP_CONCAT approach, which would output the exact meta_ids concatenated and already grouped by post_ids:
```
select GROUP_CONCAT(meta_id) as meta_ids 
from bkp_wp_postmeta wpm
where wpm.post_id in (
	select wpm2.post_id
	from bkp_wp_postmeta wpm2
	group by wpm2.post_id, wpm2.meta_key
	having count(*) > 1
)
group by wpm.post_id 
order by wpm.post_id, meta_id;
```
| meta_ids |
|------------- |
| 1,3 |
| 4,5,6 |

so we could explode delete 1, 4 and 5, leaving the newest 3 and 6.

Let me also share why I felt motivated to try make the queries a bit more readable. I noticed these several things in [the second query L:261](https://github.com/Automattic/newspack-custom-content-migrator/pull/178/files#diff-89371c8d4b14fb806dcd4d00308135320cf686634ebfe42ece522c55cd02f9a1R261):
- it should probably first be checked if it's necessary to run this query at all, with something like `if ( counter != total_unique_keys )`, or else not run this query for every post_id
- also the `CONCAT(meta_key, '.', IFNULL(meta_value, 'null')) as concat_key` is, how I read it, basically a `SELECT meta_key, meta_value`, but this concatenated string is next exploded, and also imploded by PHP, which suggests that it maybe could have been a simple `SELECT meta_key, meta_value` in the first place. Of course, the concat-explode-implode does the job 100% correctly. Seems to me that just for readability and performance it might be cleaned up a bit.

That's just my thoughts, and I'd love to get @kariae's too.

src/Migrator/General/ContentFixerMigrator.php
```diff
@@ -224,6 +233,69 @@ function( $inner_content ) use ( $shortcodes ) {
...
+				}
+			}
+			$table_result['deleted_rows'] = $de_duplicated_rows;
+			$table_results[] = $table_result;
+			$eliminated_rows              += $de_duplicated_rows;
+			$progress_bar->tick();
+		}
+		$progress_bar->finish();
+
+		$results = $wpdb->get_results( "SELECT pm.meta_id, pm.post_id, p.ID FROM $wpdb->postmeta pm LEFT JOIN $wpdb->posts p ON pm.post_id = p.ID WHERE p.ID IS NULL" );
```
iuravic, line 280: Could we also output something like "Deleting postmetas with non-existent post_ids..."?

src/Migrator/General/ContentFixerMigrator.php
```diff
@@ -224,6 +233,69 @@ function( $inner_content ) use ( $shortcodes ) {
...
+		$loose_post_meta_ids = array_map( fn($result) => $result->meta_id, $results );
+		if ( ! empty( $loose_post_meta_ids ) ) {
+			$loose_post_meta_ids = implode( ',', $loose_post_meta_ids );
+			$deleted_rows = $wpdb->query( "DELETE FROM $wpdb->postmeta WHERE meta_id IN ($loose_post_meta_ids)" );
```
iuravic, line 284: @kariae would it be feasible if we used your throttled foreach loop her to would delete these in batches, to save Atomic resources?

iuravic (changes requested):

src/Migrator/General/ContentFixerMigrator.php
eddiesshop: Great observations Ivan! I agree with you. The reason I did it in this way was to be save and not delete any meta_key-meta_value combination (due to my inexperience with wordpress DB management). But if the expectation is that there should only ever by one value per unique meta_key, then we can absolutely improve this so that it's only one query that shows the duplicate meta_ids that need to be deleted.

kariae: Thank you @eddiesshop for the amazing work put into this, I'm sure it will benefit our sites a lot. Thank you @iuravic for your feedback, the details and examples were helpful to follow your idea.

I share Eddy's point in terms that WP offers the possibility to have a post meta with multiple values, the third parameter of [get_post_meta](https://developer.wordpress.org/reference/functions/get_post_meta/) confirms this, so I think it would make sense to keep all the possible values in your example for the meta `_wp_page_template ` we should keep both values `default` and `new`, otherwise how can we know which value to keep.

src/Migrator/General/ContentFixerMigrator.php
kariae: The throttle function is for reading the posts as it gets the same arguments as `get_posts`, but I agree with you that it makes sense for our use case (knowing that this will run on Atomic) to use a throttle function to delete batches of postmeta.

kariae: @eddiesshop can we add a `dry-run` flag for the command to showcase the effect of dropping the rows before running it, please?

src/Migrator/General/ContentFixerMigrator.php
iuravic: I was definitely wrong, we can have multiple post_metas with same meta_keys. So my queries are not useful.

PR #179, opened 2022-06-27
feat: created function to sync wp_term_taxonomy.count column.

eddiesshop:
Created a function which will sync the actual row counts from wp_term_relationships column to the wp_term_taxonomy.count column.
Another function in this commit is the ability to remove Tags with associated posts below a certain threshold.

### DIFF
--- a/src/Migrator/General/TaxonomyMigrator.php
+++ b/src/Migrator/General/TaxonomyMigrator.php
@@ -94,7 +94,7 @@ class TaxonomyMigrator implements InterfaceMigrator {
 			'newspack-content-migrator sync-taxonomy-count-with-real-values',
 			[ $this, 'cmd_sync_taxonomy_count_column_with_actual_values' ],
 			[
-				'shortdesc' => 'This command will compare wp_term_taxonomy.count values with actual row counts in wp_term_relationships table.',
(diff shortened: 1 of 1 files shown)


iuravic (approved):

src/Migrator/General/TaxonomyMigrator.php
```diff
@@ -88,6 +89,48 @@ public function register_commands() {
 				],
 			],
 		] );
+
+		WP_CLI::add_command(
+			'newspack-content-migrator sync-taxonomy-count-with-real-values',
+			[ $this, 'cmd_sync_taxonomy_count_column_with_actual_values' ],
+			[
+				'shortdesc' => 'This command will compare wp_term_taxonomy.count values with actual row counts in wp_term_relationships table.',
```
iuravic, line 97: Should we update this from "will compare" to "will fix"?

iuravic (changes requested):

iuravic:
@eddiesshop , I accepted the PR, then checked back in, and added a note that we just update the command description to be a bit more accurate.

src/Migrator/General/TaxonomyMigrator.php
eddiesshop: I think I put "will compare" because the default behavior is not to update the DB table. One must pass the --update flag for this command to then fix/update the counts. I can update this though to say: "will compare and sync"

PR #180, opened 2022-07-07
Add/noozhawk migrator

kariae:
This is Noozhawk migrator, it contains the migration commands for both the old and new XML format we received from the publisher. Please check the commands for the new format only as this is what we're going to use. The old commands are for archive purposes.

The commands to check are:
```
wp newspack-content-migrator noozhawk-set-primary-category-from-meta
wp newspack-content-migrator noozhawk-set-coauthors
wp newspack-content-migrator noozhawk-import-featured-image
wp newspack-content-migrator noozhawk-migrate-galleries
```


iuravic (approved):

iuravic:
Merging while @kariae is AFK.

PR #181, opened 2022-07-07
Feat/massterlist migrator

kariae:
This is a migrator for Massterlist Editions and Job Listings. To use it you need the tables from the editions and jobs databases of the publisher, that you need to import locally, prefix the editions tables with `editions_`, and the jobs tables with `_jobs`, and then import them to the WP site where you're going to execute the commands.


iuravic:
> This is a migrator for Massterlist Editions and Job Listings. To use it you need the tables from the editions and jobs databases of the publisher, that you need to import locally, prefix the editions tables with `editions_`, and the jobs tables with `_jobs`, and then import them to the WP site where you're going to execute the commands.

Thanks for this info, @kariae 👍 This looks like something very useful to have in the command description so that we know how to run it. I'll quickly add it as a commit right now.

iuravic (approved):

iuravic:
Merging while @kariae is AFK.

PR #182, opened 2022-07-10
Feat/merge co authors

kariae:
This adds a command to merge co-authors, as often we migrate the bylines for the same authors with a typo on the name.

The command can be used as follow:
```
wp newspack-content-migrator co-authors-merge-coauthors --guest-author-id=123 --guest-authors-to-remove-ids=456,789
```

Where the co-authors with the IDs `456`, and `789` are going to be removed, and their content reassigned to the co-author with the ID `123`.

### DIFF
--- a/src/Migrator/General/CoAuthorPlusMigrator.php
+++ b/src/Migrator/General/CoAuthorPlusMigrator.php
@@ -40,7 +40,7 @@ class CoAuthorPlusMigrator implements InterfaceMigrator {
 	 */
 	private function __construct() {
 		$this->coauthorsplus_logic = new CoAuthorPlus();
-		$this->posts_logic = new Posts();
+		$this->posts_logic         = new Posts();
 	}
 
 	/**
@@ -51,8 +51,7 @@ class CoAuthorPlusMigrator implements InterfaceMigrator {
 	public static function get_instance() {
 		$class = get_called_class();
 		if ( null === self::$instance ) {
-			self::$instance = new $class;
-
+			self::$instance = new $class();
 		}
 
 		return self::$instance;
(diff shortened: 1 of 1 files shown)


src/Migrator/General/CoAuthorPlusMigrator.php
```diff
@@ -737,36 +786,74 @@ public function cmd_split_guest_author_to_multiple_authors( $args, $assoc_args )
...
+		}
+
+		foreach ( $co_authors_to_remove_ids as $co_author_to_remove_id ) {
+			$co_author_to_remove = $this->coauthorsplus_logic->get_guest_author_by_id( $co_author_to_remove_id );
+
+			if ( ! $co_author_to_remove ) {
+				WP_CLI::error( sprintf( 'There is no co-author with this ID %d.', $co_author_to_remove_id ) );
+			}
+
+			$result = $this->coauthorsplus_logic->coauthors_guest_authors->delete( $co_author_to_remove_id, $co_author->user_login );
```
iuravic, line 848: Could we please double check if this `delete` will indeed reassign from `$co_author_to_remove_id` to `$co_author`?

Looking at the [`delete` method](https://github.com/Automattic/Co-Authors-Plus/blob/master/php/class-coauthors-guest-authors.php#L1325), it seems to me that it reassigns from GA to it's linked WP User, not to the other GA target user which we specified with [`--guest-author-id`](https://github.com/Automattic/newspack-custom-content-migrator/pull/182/files#diff-7972f9c14785a69fc77578f71d0e8a020605a9806d49655e5f265fa16fff6e5eR282).

I might be wrong, but if we could just double check this please. Thanks!

iuravic (changes requested):

PR #183, opened 2022-07-13
A fix for Content Diff updating of Attachment IDs

iuravic:
Content importer should only update the newly imported Posts' att.ID, not modify the existing Posts' attachment IDs. This PR fixes that.

This bug wasn't visible previously, because it only affects external sites which use same attachment IDs that also exist on the Staging site. An example such as this is most apparent in completely separate content archives, like additional archive sites with completely separate contents, but would not be visible if a preexisting content on Staging (like Newspack pages) didn't use the same IDs, and if it didn't use featured images (attachment IDs).

An example:
- if a Staging has an existing Post 111 with attachment 123,
- and the external archive site ("live DB") has another Post 33333 with same attachment ID 123 (but this is a different attachment file), once this Post gets imported, both the Post ID will change, e.g. 44444, and the attachment ID will change too, e.g. 456
- when updating attachment ID 123 to 456, we should only do so for the newly imported post 44444, but not for the old existing post 123
- and this PR fixes that

### DIFF
--- a/src/MigrationLogic/ContentDiffMigrator.php
+++ b/src/MigrationLogic/ContentDiffMigrator.php
@@ -495,15 +495,22 @@ class ContentDiffMigrator {
 	/**
 	 * Updates Posts' Thumbnail IDs with new Thumbnail IDs after insertion.
 	 *
+	 * @param array  $imported_post_ids_map   Keys are Post IDs on Live Site, values are Post IDs of imported posts on Local Site.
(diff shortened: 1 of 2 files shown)


iuravic:
@Automattic/newspack-launch-and-infrastructure this bug is a bit urgent to fix, so I'm now merging it (since most of our team is either on AFK, or just about to leave for AFK), so please review in your own time, and we can introduce additional changes via a new PR if needed.

PR #197, opened 2022-08-29
Moves files out of those custom S3 integration subfolders in uploads

iuravic:
Some S3 integration plugins use custom subfolders for images, e.g. 
```
uploads/YYYY/MM/CUSTOM_SUBFOLDER/img.jpg.
```

This commands moves the files down a level to 
```
uploads/YYYY/MM/img.jpg
```

This is refactored from [CharlestonCityPaperMigrator](https://github.com/Automattic/newspack-custom-content-migrator/blob/master/src/Migrator/PublisherSpecific/CharlestonCityPaperMigrator.php#L82) from a year ago, and is an initial commit for the command.

Future improvements:
- the command only moves the files out of the subfolders down a level below into the MM folder, but keeps the custom subfolders -- those can be deleted after the move

### DIFF
--- a/src/Migrator/General/S3UploadsMigrator.php
+++ b/src/Migrator/General/S3UploadsMigrator.php
@@ -112,6 +112,13 @@ class S3UploadsMigrator implements InterfaceMigrator {
 				],
 			]
 		);
+		WP_CLI::add_command(
+			'newspack-content-migrator s3uploads-remove-subfolders-from-uploadsyyyymm',
+			[ $this, 'cmd_remove_subfolders_from_uploadsyyyymm' ],
+			[
+				'shortdesc' => "Some S3 integration plugin use a custom subfolder in e.g. '~/wp-content/uploads/YYYY/MM/SUBFOLDER/image.jpg'. This command removes all such subfolders, and moves all the images one level below to '~/wp-content/uploads/YYYY/MM/image.jpg'.",
+			]
+		);
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


src/Migrator/General/S3UploadsMigrator.php
```diff
@@ -354,6 +361,87 @@ public function is_s3_uploads_plugin_active() {
...
+
+					$msg = '{SUBDIRPATH}' . $mm_subdir;
+					$this->log( $log_filename, $msg );
+					foreach ( $subdir_files as $subdir_file ) {
+						$this->log( $log_filename, $subdir_file );
+						$old_file = $mm_subdir . '/' . $subdir_file;
+						$new_file = $mm_full_path . '/' . $subdir_file;
+						$renamed  = rename( $old_file, $new_file );
+						if ( false === $renamed ) {
+							$this->log( 'ccp_subfoldersMoveError', $old_file . ' ' . $new_file );
```
abdelmalekkkkk, line 434: It looks like this `log` is not using the new `$log_filename` variable for the log file. It should probably look something like `$this->log( $log_filename, $old_file . ' ' . $new_file );`

abdelmalekkkkk (changes requested):
Looks great. Just noticed one small thing with the logs.

src/Migrator/General/S3UploadsMigrator.php
iuravic: Great catch, thanks! Updated now.

abdelmalekkkkk (approved):
Everything looks good to me. Thanks!

PR #199, opened 2022-08-30
Add migrator for Lkldnow + Command for migrating avatars

abdelmalekkkkk:
I created a new migrator for Lkldnow under Migrators/PublisherSpecific called LkldNowMigrator. I also added a command that migrates the users' avatars to Simple Local Avatars. It's related to this [ticket](https://app.asana.com/0/1201792624707312/1202733251254639).

### DIFF
--- a/src/Migrator/PublisherSpecific/LkldNowMigrator.php
+++ b/src/Migrator/PublisherSpecific/LkldNowMigrator.php
@@ -1,0 +1,179 @@
+<?php
+
+namespace NewspackCustomContentMigrator\Migrator\PublisherSpecific;
+
+use \NewspackCustomContentMigrator\Migrator\InterfaceMigrator;
+use Simple_Local_Avatars;
+use \WP_CLI;
+
+/**
+ * Custom migration scripts for LkldNow.
+ */
+class LkldNowMigrator implements InterfaceMigrator {
+
+	/**
+	 * @var null|InterfaceMigrator Instance.
+	 */
+	private static $instance = null;
+
+	/**
+	 * @var null|Simple_Local_Avatars Instance of Simple_Local_Avatars
+	 */
+	private $simple_local_avatars;
+
+	/**
+	 * Constructor.
+	 */
+	private function __construct() {
+		$plugins_path = defined( 'WP_PLUGIN_DIR' ) ? WP_PLUGIN_DIR : ABSPATH . 'wp-content/plugins';
+
+		$simple_local_avatars_plugin_file = $plugins_path . '/simple-local-avatars/simple-local-avatars.php';
+		
+		if ( is_file( $simple_local_avatars_plugin_file ) && include_once $simple_local_avatars_plugin_file ) {
+			$this->simple_local_avatars = new Simple_Local_Avatars();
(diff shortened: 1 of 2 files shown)


src/Migrator/PublisherSpecific/LkldNowMigrator.php
```diff
@@ -0,0 +1,179 @@
...
+			]
+		);
+	}
+
+	/**
+	 * Extract the old avatars from meta and migrate them to Simple Local Avatars.
+	 */
+	public function cmd_lkldnow_migrate_avatars() {
+		if ( $this->simple_local_avatars == null ) {
+			WP_CLI::warning( 'Simple Local Avatars not found. Install and activate it before using this command.' );
```
iuravic, line 70: Excellent.

WP_CLI::error() can also be convenient, because it also exits and breaks execution.

src/Migrator/PublisherSpecific/LkldNowMigrator.php
```diff
@@ -0,0 +1,179 @@
...
+			if ( isset( $avatar_data['media_id'] ) ) {
+				$avatar_id = $avatar_data['media_id'];
+			} else if ( isset( $avatar_data['full'] ) ) {
+				$avatar_id = attachment_url_to_postid( $avatar_data['full'] );
+				
+				// Sometimes the avatar is uploaded without being linked to an attachment
+				// in that case we insert a new attachment
+				if ( $avatar_id == 0 ) {
+					$avatar_url = $avatar_data['full'];
+					$avatar_id = $this->assign_upload_file_to_attachment( $avatar_url );
```
iuravic, line 118: This is good, but for next time, I strongly recommend using [`MigrationLogic\Attachments::import_external_file`](https://github.com/Automattic/newspack-custom-content-migrator/blob/master/src/MigrationLogic/Attachments.php#L42). It is a reusable function we've built which does a bunch of things:
- can import an external URL, or a local file image
- can set all the meta data (title, caption, alt, etc.)

Also, do take a look at all the existing [src/MigrationLogic](https://github.com/Automattic/newspack-custom-content-migrator/tree/master/src/MigrationLogic) and [src/Migrator/General](https://github.com/Automattic/newspack-custom-content-migrator/tree/master/src/Migrator/General).

iuravic (approved):

src/Migrator/PublisherSpecific/LkldNowMigrator.php
abdelmalekkkkk: @iuravic Thanks for the reminder. I'll make sure I use that in the future.

src/Migrator/PublisherSpecific/LkldNowMigrator.php
abdelmalekkkkk: @iuravic I actually checked the Attachments class for any functions that I can use. The import_external_file is not appropriate here, since the image is already in the uploads folder, it's just not assigned to any attachment post.

abdelmalekkkkk:
@iuravic Thanks for the review. Migrating now.

PR #202, opened 2022-08-31
feat(attachment): add command to verify broken images URLs

kariae:
This PR adds a command to use the Attachments utility method `get_broken_attachment_urls_from_posts` to get broken in-story image URLs.

The method was not checking [the responsive images](https://developer.mozilla.org/en-US/docs/Learn/HTML/Multimedia_and_embedding/Responsive_images) using the`srcset` attribute from the `` tag, this is fixed now.

The command can be used in batches to run it multiple times in parallel when the number of posts to check is high. It also accepts an `index` parameter in case the command is killed and you need to restart it from the last index.

I added a parameter to the `get_broken_attachment_urls_from_posts` method called `$logger` which is a callback function that we can use while looping over the posts, a use case for it, is to pass a function that will log the missing images in the post while looping over all the posts without waiting for the loop to end.

### DIFF
--- a/src/MigrationLogic/Attachments.php
+++ b/src/MigrationLogic/Attachments.php
@@ -118,16 +118,10 @@ class Attachments {
 				continue;
 			}
 			WP_CLI::line( sprintf( 'Checking Post(%d/%d): %d', $index + 1, $total_posts, $post->ID ) );
-			$image_urls_to_check = [];
-			$broken_post_images  = [];
-
-			preg_match_all( '/<img[^>]+(?:src|data-orig-file)="([^">]+)"/', $post->post_content, $image_sources_match );
-			if ( array_key_exists( 1, $image_sources_match ) ) {
-				$image_urls_to_check = $image_sources_match[1];
-			}
+			$broken_post_images = [];
 
 			// get responsive images.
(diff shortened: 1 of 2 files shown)


src/MigrationLogic/Attachments.php
```diff
@@ -148,22 +173,55 @@ public function get_broken_attachment_urls_from_posts( $post_ids = [], $is_hoste
...
 		return $broken_images;
 	}
+
+	/**
+	 * Retreive all the image sources from the srcset attribute in post content.
+	 *
+	 * @param string $post_content post content to look for the images sources on.
+	 * @return array
+	 */
+	private function get_responsive_images_sources_from_content( $post_content ) {
```
iuravic, line 210: Could we make this function public? It might be useful for some other Migrators in the future to access too.

iuravic (changes requested):
Everything looks great! Amazing job, and an amazing script! If we could just make that one private function public, that's all the comments I have.

src/MigrationLogic/Attachments.php
kariae: Yup, that's a good idea. Actually, this method gets only the responsive images from the post content, I refactored the code a bit to get all the images from the post content using this method (using the `src` attribute and the responsive images), that works for you?

iuravic (approved):

PR #205, opened 2022-09-01
Add command to migrate republished content from meta to paragraphs

abdelmalekkkkk:
I created this command for this [ticket](https://app.asana.com/0/1201792624707312/1202664404571760). All it does is take the original's content link and name from the 'aalink' and 'aaname' custom fields, and add a paragraph to the end of each article with a link to that original article.

I also made the script `sleep` every 100 posts to avoid any issues with the database.

### DIFF
--- a/src/Migrator/PublisherSpecific/LkldNowMigrator.php
+++ b/src/Migrator/PublisherSpecific/LkldNowMigrator.php
@@ -5,6 +5,7 @@ namespace NewspackCustomContentMigrator\Migrator\PublisherSpecific;
 use \NewspackCustomContentMigrator\Migrator\InterfaceMigrator;
 use Simple_Local_Avatars;
 use \WP_CLI;
+use WP_Query;
 
 /**
  * Custom migration scripts for LkldNow.
@@ -57,6 +58,14 @@ class LkldNowMigrator implements InterfaceMigrator {
 			[ $this, 'cmd_lkldnow_migrate_avatars' ],
 			[
 				'shortdesc' => 'Migrates the users\' avatars from WP User Avatars to Simple Local Avatars.',
+				'synopsis'  => [],
+			]
+		);
+		WP_CLI::add_command(
+			'newspack-content-migrator lkldnow-migrate-republished-content',
+			[ $this, 'cmd_lkldnow_republished_content' ],
+			[
+				'shortdesc' => 'Append a hyperlink to the original article to the end of the article.',
 				'synopsis'  => [],
 			]
 		);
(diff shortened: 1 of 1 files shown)


src/Migrator/PublisherSpecific/LkldNowMigrator.php
```diff
@@ -162,6 +171,95 @@ public function cmd_lkldnow_migrate_avatars() {
...
+		);
+
+		$query = new WP_Query( $args );
+
+		$updated_posts_count = 0;
+
+		foreach ( $query->posts as $index => $post ) {
+			if ( ( $index  + 1 ) % 100 == 0 ) {
+				WP_CLI::log( 'Sleeping for 1 second... ');
+				sleep( 1 );
```
iuravic, line 213: Was the reason behind this sleep to minimize MySQL impact on Atomic? If so, thanks for thinking about this, and for being se thoughtful!

However I think you can feel free to skip this sleep here. We'll mostly want to sleeps when we do raw MySQL queries, and should split them with offsets & limits, too.

No "request changes" here are necessary -- if you're able to migrate the content with this sleep amount, feel free to continue. If you find it's taking too long, just comment it out.

src/Migrator/PublisherSpecific/LkldNowMigrator.php
```diff
@@ -162,6 +171,95 @@ public function cmd_lkldnow_migrate_avatars() {
...
+		$log_filename = 'lkld_updated_articles.log';
+
+		$already_updated_meta_key = '_newspack_original_content_appended';
+
+		$source_link_meta_key = 'aalink';
+		$source_name_meta_key = 'aaname';
+
+		$paragraph_template = '
+		
+		<!-- wp:paragraph -->
```
iuravic, line 188: @abdelmalekkkkk , make sure you test these in real life, too.

From my experience, Gutenberg can be very sensitive to extremely simple things like extra spaces, like the tabs at the beginning of 		 ..." here. If it works, that's great, and no issues, just sharing my previous experience 👍

src/Migrator/PublisherSpecific/LkldNowMigrator.php
```diff
@@ -162,6 +171,95 @@ public function cmd_lkldnow_migrate_avatars() {
...
+				esc_url( $source_link ),
+				esc_html( $source_name ),
+			);
+
+			$new_post_content = $post->post_content . $formatted_paragraph;	
+
+			$this->log( $log_filename, sprintf( "New content for post #%d:\n%s", $post->ID, $new_post_content ) );
+
+			if ( ! $dry_run ) {				
+				$result = wp_update_post( array(
```
iuravic, line 245: I was once told that $wpdb->update() is quite faster than this. If you'll like, do give it a try in some of your next migrators. P.S. $wpdb->update() can also use a wp_cache_flush() once it's done running.

iuravic (approved):
Looks great! I added some comments here too, but those are not "requested changes", just sharing my thoughts and advices :)

src/Migrator/PublisherSpecific/LkldNowMigrator.php
abdelmalekkkkk: Correct. There weren't many posts so the sleep didn't cause much of an issue. Better late than sorry 😁

abdelmalekkkkk:
Your comments are very helpful. Thanks!

PR #206, opened 2022-09-06
feat(Attachment Migrator): add command to delete in-story posts attachments

kariae:
This PR adds the command `newspack-content-migrator attachments-delete-posts-images` that can be used to delete all attachments that are used in post content.

The delete option should be done very cautiously because attachments caused in post content can also be used in page content, other post types content (e.g. newsletters, listings, ...). or within other plugins. 

The command [skips the attachments used in all the post types except the `post` post type](https://github.com/Automattic/newspack-custom-content-migrator/blob/feat/command-to-delete-posts-attachments/src/Migrator/General/AttachmentsMigrator.php#L430). It skips also the [themes logos attachments](https://github.com/Automattic/newspack-custom-content-migrator/blob/feat/command-to-delete-posts-attachments/src/Migrator/General/AttachmentsMigrator.php#L452), [attachments used in custom CSS](https://github.com/Automattic/newspack-custom-content-migrator/blob/feat/command-to-delete-posts-attachments/src/Migrator/General/AttachmentsMigrator.php#L479), [co-authors avatars](https://github.com/Automattic/newspack-custom-content-migrator/blob/feat/command-to-delete-posts-attachments/src/Migrator/General/AttachmentsMigrator.php#L509), [avatars from Local Avatars plugin](https://github.com/Automattic/newspack-custom-content-migrator/blob/feat/command-to-delete-posts-attachments/src/Migrator/General/AttachmentsMigrator.php#L528), and [Yoast default featured image](https://github.com/Automattic/newspack-custom-content-migrator/blob/feat/command-to-delete-posts-attachments/src/Migrator/General/AttachmentsMigrator.php#L548).

### Usage
- Run the command the first time without the `--confirm-deletion` flag (the other flags can be used), this will mark the attachments to be deleted with the `_newspack_attachment_to_delete` post meta, and will move the attachment file to a trash folder.
- After the QA is done, and making sure we didn't mark an attachment file to be deleted that is supposed to not be deleted, we'll run the command a second time with the `--confirm-deletion` flag, this one is destructive if not used with the `--dry-run` flag, it will delete the attachment posts marked from the first step, and delete their files from the trash folder.

### Command Flags
- `dry-run` Will show the command output without doing any action.
- `confirm-deletion` is used in the second command call after QA'ing the results from the first call, to permanently delete the attachments from the database and the files system.
- `skip-from` and `skip-to` will skip deleting attachments created in this period, can be useful if we know the Newspackification period, so we can skip the attachments used on it.

### Tests
I tested the command on a publisher site that I cloned locally, you can do the same, or create a fresh Newspack site locally, add a few listings, Newsletters, pages, and posts all with attachments, install CAP, and set some co-authors avatars. Do the first run without the `--confirm-deletion` flag and a second run with the flag, and check that only the attachments used in the posts are deleted.


src/Migrator/General/AttachmentsMigrator.php
```diff
@@ -285,6 +327,307 @@ public function cmd_switch_local_images_urls_to_s3_urls( $pos_args, $assoc_args
...
+
+				if ( ! $attachment_file ) {
+					$this->log( self::DELETING_MEDIA_LOGS, sprintf( 'Skipping deleting media (%d/%d): %d Can\'t locate its file.', $index + 1, $total_attachment_posts, $attachment_post_to_be_deleted->ID ) );
+					continue;
+				}
+
+				$media_path = $this->get_trash_folder() . '/' . $attachment_file;
+				if ( file_exists( $media_path ) ) {
+					unlink( $media_path );
+					$this->log( self::DELETING_MEDIA_LOGS, sprintf( 'Delting media (%d/%d): %d (%s)', $index + 1, $total_attachment_posts, $attachment_post_to_be_deleted->ID, $media_path ) );
```
eddiesshop, line 379: Don't know if you care about this, but 'Deleting' is misspelled here.

src/Migrator/General/AttachmentsMigrator.php
```diff
@@ -285,6 +327,307 @@ public function cmd_switch_local_images_urls_to_s3_urls( $pos_args, $assoc_args
...
+				} else {
+					$this->log( self::DELETING_MEDIA_LOGS, sprintf( 'Media File not in trash folder (%d/%d): %d (%s)', $index + 1, $total_attachment_posts, $attachment_post_to_be_deleted->ID, $media_path ) );
```
eddiesshop, line 381: How about, "File not exists, media file not in trash folder..." instead?

src/Migrator/General/AttachmentsMigrator.php
```diff
@@ -285,6 +327,307 @@ public function cmd_switch_local_images_urls_to_s3_urls( $pos_args, $assoc_args
...
+							$this->log( self::DELETING_MEDIA_LOGS, sprintf( 'Warning: Media file not located (%d/%d): %d (%s)', $index + 1, $total_attachment_posts, $attachment_post->ID, $media_path ) );
+						}
+					}
+
+					$this->log( self::DELETING_MEDIA_LOGS, sprintf( 'Marking media to be deleted (%d/%d): %d (%s)', $index + 1, $total_attachment_posts, $attachment_post->ID, $media_path ) );
+				}
+			}
+		}
+
+		wp_cache_flush();
```
eddiesshop, line 463: OoOoo TIL!

eddiesshop (approved):

src/Migrator/General/AttachmentsMigrator.php
kariae: Yeah, so you don't forget to run it after running the command :D

PR #210, opened 2022-09-15
Diff of local files VS files on S3

iuravic:


iuravic:
Abandoned in favor of https://github.com/Automattic/newspack-custom-content-migrator/pull/240

PR #211, opened 2022-09-16
Add reusable Simple Local Avatars logic

abdelmalekkkkk:


src/MigrationLogic/SimpleLocalAvatars.php
```diff
@@ -0,0 +1,58 @@
...
+	 * Attach an avatar to a user through Simple Local Avatars
+     * 
+     * @param int $user_id The user ID
+     * @param int $attachment_id The attachment ID that has the avatar
+     * @param string $rating The avatar rating (G, PG, R, X)
+	 *
+	 * @return boolean True on success, false otherwise
+	 */
+	public function import_avatar( $user_id, $attachment_id, $rating ) {
+        $this->simple_local_avatars->assign_new_user_avatar( (int) $attachment_id, $user_id );
```
iuravic, line 46: Could you run PHPCS on this file? We at least want the indentations to be the same. Thanks!

As a rule, we need to run PHPCS on all general and reusable code, and you don't have to run it on publisher specific ones.
But please still set the indentations in your IDE to use tabs -- this is what the whole A8C WP community uses and that's why we have tabs in our Sniffer standards.

src/MigrationLogic/SimpleLocalAvatars.php
```diff
@@ -0,0 +1,58 @@
...
+
+    /**
+	 * Constructor.
+	 */
+	public function __construct() {
+		$plugins_path = defined( 'WP_PLUGIN_DIR' ) ? WP_PLUGIN_DIR : ABSPATH . 'wp-content/plugins';
+
+		$simple_local_avatars_plugin_file = $plugins_path . '/simple-local-avatars/simple-local-avatars.php';
+		
+		if ( is_file( $simple_local_avatars_plugin_file ) && include_once $simple_local_avatars_plugin_file ) {
```
iuravic, line 31: Could we perhaps add a simple method to make it easy for us to check if SLA is on?
For example:
```
public function is_sla_plugin_active() {
    return $this->sla_logic->simple_local_avatars != null;
}
```

The command below runs the method `cmd_lkldnow_migrate_avatars` which checks for it like this
```
if ( $this->sla_logic->simple_local_avatars == null ) {
```
but if you're using `$this->sla_logic` for the first time, it would be much easier if there was a public method `is_sla_plugin_active` instead of reading its code and figuring out that the constructor might leave it null if the plugin file is not found.

iuravic (changes requested):

iuravic:
Added the whole Launch Team as reviewers.

src/MigrationLogic/SimpleLocalAvatars.php
abdelmalekkkkk: My bad. My editor switched to using spaces for some reason. I fixed that.

src/MigrationLogic/SimpleLocalAvatars.php
abdelmalekkkkk: Good point. I added the is_sla_plugin_active method. Thanks!

src/MigrationLogic/SimpleLocalAvatars.php
```diff
@@ -0,0 +1,62 @@
...
+		$avatar = get_user_meta( $user_id, self::AVATAR_META_KEY, true );
+
+		if ( ! empty( $avatar ) ) {
+			return false;
+		}
+
+		return true;
+	}
+
+	public function is_sla_plugin_active() {
```
iuravic, line 59: @abdelmalekkkkk the PHPCS still reports errors on this file. Here's how you can run PHPCS:

image

Ideally good IDE integration should fix these automatically. But until you set that up, follow these errors one by one, and fix them manually until they're all gone.

Also running `./vendor/bin/phpcbf --standard=phpcs.xml src/MigrationLogic/SimpleLocalAvatars.php` will fix all the errors automatically which can be fixed that way. The remaining errors you need to fix automatically.

And we should be doing this on all reusable code.

iuravic (changes requested):

src/MigrationLogic/SimpleLocalAvatars.php
abdelmalekkkkk: Thank for pointing this out. I think I messed up the PHPDocs enough to never do it again 😄

Fixed!

abdelmalekkkkk:
@iuravic Thanks for the review. I made the requested changes and merged this to master, since I need it for [this Baltimore task](https://app.asana.com/0/1201792624707312/1202703102746936).

PR #215, opened 2022-09-21
Add missing function readline

iuravic:
\readline() seems to have gone missing from Atomic recently. This adds it back to this migrator so we can use it.

@abdelmalekkkkk , I'm not a 100% sure if you were researching this same functionality as \readline(), but perhaps we could add this code somewhere more reusable, if it would be of use to other Migrators too.


abdelmalekkkkk:
Yes, I did something similar for the PrelaunchSiteQAMigrator.

image

Adding it somewhere to make it reusable is a great idea. Maybe we need a new file for helper functions, since this doesn't seem to belong to any class?

PR #225, opened 2022-11-17
GiveWP exporter

iuravic:
The unfinished GiveWP exporter.

GiveWP has a built in Export/Import feature in Dashboard, which should be used instead of this. This was made for a Publisher whose GiveWP license has expired and we needed to pull out data manually. However the migrator wasn't fully finished. I'm still sharing the code with our Team and our codebase just in case someone might find it useful in the future.


iuravic:
Abandoned in favor of built in exporter and importer and a better new workflow.

PR #226, opened 2022-11-17
Avlwatchdog donations CSV conversion

iuravic:
Created for Avlwatchdog, hence the branch name.

This is just a helper plugin to enable preparing data for https://github.com/woocommerce/woocommerce-subscriptions-importer-exporter . It also serves like a recap boilerplate reminder of how to use the importer plugin, for condensed knowledge sharing and reusability efficiency.

The code is not reusable now, which is why an exception is thrown at the beginning. The code was made to convert one type of CSV/TSC we had previously prepared for [customerorder-csv-import-suite](https://woocommerce.com/products/customerorder-csv-import-suite/?quid=9f61b4c70a0d9553175226131c78188d) donations CSV, to a CSV to be used by the [woocommerce-subscriptions-importer-exporter](https://github.com/woocommerce/woocommerce-subscriptions-importer-exporter) which imports Subscriptions. So that's a demo of how a CSV could and should be prepared and we should generalize it the next time we use it.


iuravic:
Publisher launched, code not reusable.

Issue #200, opened 2022-08-30
Check for and import revised content


gamebits:
Publishers may be editing content, no matter how old, for any number of professional, ethical, or legal reasons. And we can't count on one primary point of contact knowing the breadth of these changes: in the time between Berkeleyside's last import (Feb 2022) and their launch (July 2022), previously imported content from over sixty (60!) different authors had been modified in their production environment.

There are two steps we can use during the content refresh process to ensure that revisions to legacy content are migrated:

1. Currently, if a post has had its title, `post_name`, status, or date changed (most likely the title), then the content migrator sees it as a new post, even if it was previously migrated. In those scenarios, if `post_name` hasn't changed, then the result may be posts with duplicate slugs. We'll want the one with the most recent `post_modified` timestamp to overwrite the older one.
2. Currently, if only the `post_content` has changed, then it will _not_ be seen as a new post and will not be migrated. We should still run a comparison of all previously migrated posts to see if their counterparts still have matching `post_modified` values; if not, then the staging site's post should be updated with the new `post_content` and `post_modified` values.

More details in P2: pamTN9-4Tp-p2

gamebits:
Also being tracked in Asana: 1202419618513283-as-1202904873161609

PR #236, opened 2022-12-16
Content Diff refresh modified posts

iuravic:
Checks if any of the posts on Staging were modified on Live with a greater `post_modified` date, and it then reimports them.

How to test:
- mysqldump of "live site" which has a few posts that have `post_modified` > `post_modified` on Staging; import those tables with `live_` prefix side by side to the local "staging" DB
- run the `newspack-content-migrator content-diff-search-new-content-on-live` command and check that the new `content-diff__modified-ids.log` gets created with those IDs
- run the regular Content Diff import command, and check that the Posts have been updated


gamebits:
This will close https://github.com/Automattic/newspack-custom-content-migrator/issues/200.

PR #237, opened 2022-12-19
Add checking for raw shortcodes to QA migrator

abdelmalekkkkk:


src/Utils/Logger.php
```diff
@@ -20,6 +20,10 @@ class Logger {
 	 * @return string Directory path.
 	 */
 	public function get_le_log_path() {
+		$path = get_temp_dir() . self::LE_LOG_DIRECTORY;
+		if ( ! file_exists( $path ) ) {
+			mkdir( $path );
+		}
 		return get_temp_dir() . self::LE_LOG_DIRECTORY;
```
iuravic, line 27: Let's return `$path` instead, since this var now exists.

src/Logic/Posts.php
```diff
@@ -464,4 +464,50 @@ public function delete_post_revisions( $post_id, $chunk_size ) {
...
 	}
+
+	/**
+	 * Find all posts with raw shortcodes
+	 * 
+	 * @param callback $logger Function to call for logging.
+	 * 
+	 * @return array Associative array where ID is the Post ID and the value is an array of shortcodes used
+	 */
+	public function find_shortcodes_in_posts( $logger = false ) {
```
iuravic, line 475: Let's move these two methods to a new migrator `NewspackCustomContentMigrator\Logic\Shortcodes`

iuravic (changes requested):
Looks great 👍 

Just a couple of things
- let's use a new Shortcodes migrator class
- that one variable return is a tiny detail, not crucial
- please run PHPCS and fix errors. You'll find there are old existing errors in `src/Command/General/PostsMigrator.php` too, would be great to fix as much of those you can

PR #241, opened 2022-12-27
Fix for "sync-taxonomy-count-with-real-values" command

iuravic:
The command "sync-taxonomy-count-with-real-values" did not fix counts for categories which should have zero count.

Fixed and rewrote the command from using a complex DB query to use the built in WP `wp_update_term_count_now()`.

Renamed command `sync-taxonomy-count-with-real-values` to `fix-category-and-tag-count`.
Added a new command `fix-taxonomy-count` which can fix any other taxonomy count.

Also ran PHPCS and fixed some things, but not all of them.


iuravic:
@eddiesshop tagging you since this is a fix for the previous "sync-taxonomy-count-with-real-values" command

kariae (approved):

PR #243, opened 2022-12-31
Convert WP author users to GAs

iuravic:
A new reusable command creates GAs from WP author users and assigns them to all or just specific posts.


iuravic:
Closing as out of date.

PR #259, opened 2023-02-05
feat(MetroMigrator): add commands to fix missing jpe images

kariae:
_By @kariae , Feb 5 2023_:
Add two commands to the Metro Migrator. The first one is to fix the attachment posts with `.jpg` extension and change their extensions from `.jpe` to `.jpeg`. 

The second command checks over the posts and pages' content and rename the attachments with `.jpe` to `.jpeg` if their files exist.

---

_By @iuravic , Jul 28 2025_:
Reopened due to an ongoing Metro CMS migration:
- rebased and updated to solve long difference in history from this branch to trunk
- apparently Metro version was updated with several data changes, property types and values changes
    - "tag" object's property changed from "describes" to "default"
    - "tag" object's "type" property now has possible values "default" or "person"
    - author "tag" object was changed from "authored" to "person"
- txt mappings files were no longer successfully created, general tags and authors creation and operations did not work
- modified the $tag_exists to query against DB instead agains mappings
- cmd_metro_find_tags_types fixed how it iterate's over the new JSON structure
- added support for multiple authors
- for unknown reason some author "tag" objects have "type" = "default" instead of expected "person" (possibly a new byline system in Metro), so now also creating missing authors from "default" tags on the fly (not just relying that they'll all be created in command which creates authors)
- multiple syntax fixes for deprecations, some extra handling added
- now using NMT logic but with a minimal intervention to make the commands work
- authors now created as Guest Contributors with identical logic for email generation with @example.com domain which was mirrored from UsersHelper
- refactored to work with new NCCM/NMT command command structure
- changed how mappings path is hardcoded to ABSPATH, now using WP_CONTENT_DIR, because ABSPATH does not work on Atomic as expected


iuravic:
Closing as out of date.

PR #265, opened 2023-02-10
Retro Report importers

philipjohn:
Includes importers for:

* Standards


src/Logic/Sponsors.php
```diff
@@ -0,0 +1,89 @@
...
+	 * @param int $post    ID of the post to be sponsored
+	 *
+	 * @return bool True is successful, false on failure.
+	 */
+	public function add_sponsor_to_post( $sponsor, $post ) {
+
+		// Make sure we have a sponsor post.
+		$sponsor_post = get_post( $sponsor );
+		if ( ! is_a( $sponsor_post, 'WP_Post' ) ) {
+			$this->logger->log( 'sponsors', sprintf( 'No sponsor found with ID %d', $sponsor ), false );
```
iuravic, line 49: Since `'sponsors'` log file name is used as a repeated string in multiple places, could we please,
- make it a `const` as well
- add a suffix to it, like `sponsors.log` or `txt` to make it a bit more obvious

Optionally but reall less important, I'd recommend adding something like this in the end `WP_CLI::log( sprintf( "Done. See %s file.", self::LOG_FILENAME ) );`. It's bit more convenient for those executing so that they don't have to read code to notice that a log was generated.

iuravic (changes requested):

PR #269, opened 2023-03-01
feat: attempting to create a standardized approach to handling common…

eddiesshop:
… export file types that we receive for importation during the migration process.

This all culminates in being able to import a CSV or JSON file with the following lines of code:

```
$iterator = ( new FileImportFactory() )
			->get_file( $file_path )
                         //->set_start( $arbitrary_start_row_numberer )
                         //->set_end( $arbitrary_end_row_number )
			->getIterator();

foreach ( $iterator as $row ) {
    // process row
}
```

### DIFF
--- a/src/Utils/AbstractFile.php
+++ b/src/Utils/AbstractFile.php
@@ -1,0 +1,112 @@
+<?php
+/**
+ * Abstract class definition of File.
+ *
+ * @package NewspackCustomContentMigrator
+ */
+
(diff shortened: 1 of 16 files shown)


iuravic:
I really like this -- it's simple and its function is clear.

Just one structural thing -- since this is basically a bundle of files, it would make sense to put all of these files into its own namespace inside Utils, something like `src/Utils/CommonDataFileIterator`, or similar if you think of a different naming.

PR #270, opened 2023-03-06
Add/saporta report migrator

kariae:
This PR introduces the Saporta Report migrator and some internal tooling improvements. Some Commits that need some code review, please:

[101438b](https://github.com/Automattic/newspack-custom-content-migrator/pull/270/commits/101438bfc163100346d7f90ef6c97042add3c0d2) Adds logging levels when we log to CLI.

[c7b9cd6](https://github.com/Automattic/newspack-custom-content-migrator/pull/270/commits/c7b9cd6c509acad46ad760a902d867a70777719f) Introduce the Gutenberg Blocks Generator class, for now, it contains two functions to generate the `Jetpack Tiled Gallery` and the `Jetpack Slideshow` blocks, and a helper function to generate the blocks PHP array to be used for the upcoming methods.

The idea is that whenever we want to add a Gutenberg Block generator method, to create this block in a post, copy its code, below is an example of the Paragraph block, pass the code through the helper method `get_block_json_array_from_content` which will result in a JSON format array that we'll pass through a tool like [Convert JSON to PHP Array](https://wtools.io/convert-json-to-php-array) to get the PHP array code and use it in the Block generator method.

### DIFF
--- a/src/Logic/GutenbergBlockGenerator.php
+++ b/src/Logic/GutenbergBlockGenerator.php
@@ -81,7 +81,7 @@ class GutenbergBlockGenerator {
 		return [
 			'blockName'    => 'jetpack/tiled-gallery',
 			'attrs'        => [
-				'columnWidths' => $tile_sizes,
+				'columnWidths' => [], // It will be filled after fixing the block manually from the backend.
 				'ids'          => $attachment_ids,
 			],
 			'innerBlocks'  => [],
--- a/src/Logic/GutenbergBlockGenerator.php
+++ b/src/Logic/GutenbergBlockGenerator.php
@@ -2,11 +2,20 @@
 /**
  * Generator for Gutenberg Blocks content.
  *
+ * To add a new method to this class that generates a Gutenberg Block, please follow these instructions:
+ *     - Create a draft post in Gutenberg and add the desired block.
+ *     - Copy the source HTML of your block (from the code editor) and pass it to the `get_block_json_array_from_content` method.
+ *     - Transform the JSON result to a PHP array using a tool like https://wtools.io/convert-json-to-php-array
+ *     - The newly created method should return this PHP array.
+ *     - Add method parameters as necessary to customize the resulting array.
+ *     - The idea is to use the output of your method as an input to the serialize_block(s) to get the HTML content of the block.
+ *
  * @package GutenbergBlockGenerator
  */
 
 namespace NewspackCustomContentMigrator\Logic;
 
+use \NewspackCustomContentMigrator\Utils\Logger;
 use \WP_CLI;
 
 /**
@@ -15,6 +24,18 @@ use \WP_CLI;
  * @package NewspackCustomContentMigrator\Logic
  */
 class GutenbergBlockGenerator {
+    /**
+	 * @var Logger.
+	 */
+	private $logger;
+
+	/**
+	 * Constructor.
+	 */
+	private function __construct() {
+		$this->logger = new Logger();
+	}
+
     /**
      * Generate a Jetpack Tiled Gallery Block.
      *
@@ -41,7 +62,7 @@ class GutenbergBlockGenerator {
 
                         if ( ! $attachment_url ) {
                             $non_existing_attachment_indexes[] = $index;
-                            WP_CLI::warning( sprintf( "Attachment %d doesn't exist!", $attachment_id ) );
+                            $this->logger->log( 'jetpack_tiled_gallery_migrator.log', sprintf( "Attachment %d doesn't exist!", $attachment_id ), Logger::WARNING );
                             return null;
                         }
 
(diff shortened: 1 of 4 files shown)


src/Logic/GutenbergBlockGenerator.php
```diff
@@ -0,0 +1,176 @@
...
+ *
+ * @package GutenbergBlockGenerator
+ */
+
+namespace NewspackCustomContentMigrator\Logic;
+
+use \WP_CLI;
+
+/**
+ * Class ContentDiffMigrator and main logic.
```
iuravic, line 13: Could we perhaps add a short bullet-list type instructions in the class comment, or somewhere else, which describes how we can create new methods, please? Something like,
"To add new block methods:
- create a draft post in Gutenberg and add the desired block
- copy the source HTML, and paste it in a new method...
- use something ... e.g. should the new method return `get_block_json_array_from_content()` in some way?
- or use the WP block functions in some of these steps...

These steps ☝️ are all wrong for sure 😅 , it would just be awesome if we had a simple reminder how to use and/or add new methods 🙇

src/Logic/GutenbergBlockGenerator.php
```diff
@@ -0,0 +1,176 @@
...
+            ' ',
+            array_filter(
+                array_map(
+                    function( $index, $attachment_id ) use ( &$tile_sizes, &$non_existing_attachment_indexes, $tile_sizes_list ) {
+                        $attachment_url = wp_get_attachment_url( $attachment_id );
+
+                        if ( ! $attachment_url ) {
+                            $non_existing_attachment_indexes[] = $index;
+                            WP_CLI::warning( sprintf( "Attachment %d doesn't exist!", $attachment_id ) );
+                            return null;
```
iuravic, line 45: Do you think it would be useful to save a list of not found IDs in a log file? Just asking for your opinion. Perhaps not, if we save the whole output when running the command? Or perhaps we could add a log file name const and just log all such errors with `$this->logger->log()`?

Same for,
- `get_jetpack_slideshow()` , "Attachment %d doesn't exist!"

src/Utils/Logger.php
```diff
@@ -26,16 +32,28 @@ public function get_le_log_path() {
...
-	 * @param string  $file File name or path.
-	 * @param string  $message Log message.
-	 * @param boolean $to_cli Whether to output the message to the CLI. Default to false.
+	 * @param string         $file File name or path.
+	 * @param string         $message Log message.
+	 * @param string|boolean $level Whether to output the message to the CLI. Default to `line` CLI level.
 	 */
-	public function log( $file, $message, $to_cli = true ) {
+	public function log( $file, $message, $level = 'line' ) {
 		$message .= "\n";
```
iuravic, line 40: This is something left over from a previous commit, but this "\n" ends up in WP_CLI::log too, and causes a double line break.

Could we please clean this up and:
- remove this line,
- then WP_CLI::log will only output this in a new line nicely
- and then add the "\n" to file_put_contents, like `file_put_contents( $file, "\n" . $message, FILE_APPEND );`

Thank you!

iuravic (changes requested):

src/Logic/GutenbergBlockGenerator.php
kariae: Yup, it makes sense, I'll add that.

src/Logic/GutenbergBlockGenerator.php
```diff
@@ -0,0 +1,197 @@
+<?php
+/**
```
iuravic, line 2: @kariae , woohoo! 🥳

src/Utils/Logger.php
```diff
@@ -26,16 +33,30 @@ public function get_le_log_path() {
...
-	 * @param boolean $to_cli Whether to output the message to the CLI. Default to false.
+	 * @param string         $file File name or path.
+	 * @param string         $message Log message.
+	 * @param string|boolean $level Whether to output the message to the CLI. Default to `line` CLI level.
 	 */
-	public function log( $file, $message, $to_cli = true ) {
-		$message .= "\n";
-		if ( $to_cli ) {
-			WP_CLI::line( $message );
+	public function log( $file, $message, $level = 'line' ) {
```
iuravic, line 40: Since we're changing the number and order of parameters here, this will break backwards compatibility with,
- Command/General/PostsMigrator, [line](https://github.com/Automattic/newspack-custom-content-migrator/blob/master/src/Command/General/PostsMigrator.php#L668)

We can update the PostsMigrator class, but I personally would vote that we kept the `$to_cli = true` param, since I sometimes like not to output to CLI for some very low-level detail logging, and in this case if we kept the existing args and just added the new $level like `log( $file, $message, $to_cli = true, $level = 'line' )` , it would also keep backwards compatibility.

iuravic (changes requested):

src/Utils/Logger.php
kariae: @iuravic sounds logical to me, I just find it redundant to have the level and the CLI, and if you pass `false` as level it won't output to the CLI, so this won't break [this line](https://github.com/Automattic/newspack-custom-content-migrator/blob/master/src/Command/General/PostsMigrator.php#L668), there is a [check before outputting to the CLI](https://github.com/Automattic/newspack-custom-content-migrator/pull/270/files#diff-44097db48bf8d8bbb3d223b2e57d01674e8f63f7ff767bdd226d72248fb14b0bR41)

iuravic: Ah, that's right, sorry I missed it.

iuravic (approved):

PR #271, opened 2023-03-07
Add/vtdigger

iuravic:
## New reusable code in this PR
- `src/Command/General/CoAuthorPlusMigrator.php`
  - `cmd_export_posts_gas` creates a simple and convenient `.php` file with array containing post ID as value, and subarray of GA display names. Handy for quickly exporting/importing GAs to posts, until our Content Diff does it automatically
- `src/Command/General/InlineFeaturedImageMigrator.php`
  - `cmd_de_dupe_featured_images` removes featured images anywhere from `post_content`, so that they're not repeated twice in post template. Searches and removes both `wp:image` blocks, and `` elements
- `src/Command/General/ReusableBlocksMigrator.php`
  - new command `cmd_delete_reusable_blocks_from_content` takes a list of reusable blocks' Post IDs and removes usages from all Posts' `post_content`, e.g. give it `401522` and it removes `<!-- wp:block {"ref":401522} /-->` from all posts
- `src/Logic/Taxonomy.php`
  - a couple of commands for more specific ops with Categories


kariae:
Thanks, Ivan for the new commands, I have a few questions regarding some of them:

In the `cmd_export_posts_gas` the result contains some posts with NULL, they seem to be posts linked to WP User authors and not co-authors, do you think we can skip adding these to the final result or is it fine to have them?

The `cmd_de_dupe_featured_images` works as expected but it does delete ALL the occurrences of the featured image in the post content which can be wrong if the article contains the featured images as part of the inner content following this format for example:
```
Paragraph
Paragraph
Featured Image
Paragraph
```

Which I don't think is what we need, am I wrong?

iuravic:
> the `cmd_export_posts_gas` the result contains some posts with NULL, they seem to be posts linked to WP User authors and not co-authors, do you think we can skip adding these to the final result or is it fine to have them?

Great catch. Yes, so this one exports just the GAs names associated to post IDs. And if a post has a WP User author but no GAs, that ID will be null.

How about for now we updated the command comment to explain this and make it even clearer -- done now in https://github.com/Automattic/newspack-custom-content-migrator/pull/271/commits/ef01f42b36543568a35986b85c52b0219b6a9b9b -- and then if we have a use case where we want to expand this functionality, we do it then?


> The cmd_de_dupe_featured_images works as expected but it does delete ALL the occurrences of the featured image in the post content... (...) ... Which I don't think is what we need, am I wrong?

I understand what you're saying.

Firstly I added a warning in command comment here https://github.com/Automattic/newspack-custom-content-migrator/pull/271/commits/a225636dcdc18df46c1f48ab3b2bc4e3fed572fb .

Next I asked this Pub's TAMs to double confirm if removing all images is really what we need for this specific publisher:

- if the answer will be yes, we can keep this command but extend it to just delete the first image by default, and only optionally delete all of them
- if the answer will be no, I'll next update this command to work differently and just delete the first image.

Thirdly, luckily this saves detailed log files, e.g. "111_before.txt" and "111_after.txt" for every post ID, so restoring them should be super simple.

kariae:
Looks good to me, thanks for the explanations. 

Can we add the warning as confirmation in the CLI maybe so it would be visible when we execute the command?

kariae (approved):

iuravic:
> Looks good to me, thanks for the explanations.
> 
> Can we add the warning as confirmation in the CLI maybe so it would be visible when we execute the command?

Now updated the reusable command to remove featured images just from the beginning of post_content. Merging now.

PR #274, opened 2023-03-17
Add/musically migrator

kariae:
This migrator comes with a bunch of methods for the GutenbergBlockGenerator that need some testing, please.


iuravic (changes requested):
Looks great 👍 

Could we just run PHPCS on src/Logic/GutenbergBlockGenerator.php please?

iuravic (approved):

iuravic (approved):

PR #279, opened 2023-04-17
feat: exit with code 1 when exceeding Atomic memory limit

kariae:
This PR adds a ticker to exit script execution with an exit code equal to 1 when reaching 0.85GB of memory. The idea is not to reach the 1GB which will cause a hard stop from the atomic environment, by doing it this way we can catch the exit code and relaunch the script with a bash script like the one below. 

```bash
#!/bin/bash

till_done() {
    $@
    exitcode=$?

    if [ $exitcode -eq "1" ]
    then
        echo "Re-running ..."
        till_done "$@"
    fi
}

till_done "$@"
```

An example of executing a WP command this way would be:

```bash
./exec_till_done.sh wp newspack-content-migrator town-news-migrate-content --export-dir-path=/tmp/content_migration/export
```

The next step is to add this bash script to the knife tool so we don't have to upload it to each site where we want to execute heavy-lifting commands.

### DIFF
--- a/newspack-custom-content-migrator.php
+++ b/newspack-custom-content-migrator.php
@@ -21,6 +21,7 @@ if ( ! defined( 'WP_CLI' ) || ! WP_CLI ) {
 
 require_once ABSPATH . 'wp-settings.php';
 
+PluginSetup::register_ticker();
 PluginSetup::setup_wordpress_importer();
 PluginSetup::register_migrators(
 	array(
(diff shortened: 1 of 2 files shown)


iuravic (approved):

iuravic:
@kariae , approved, though I believe new Atomic limit is 512 MB,
```
php -r "echo ini_get('memory_limit');"
```

PR #285, opened 2023-04-25
add/soccer-america-migrator-2

iuravic:
Second PR for SA with all changes and additions to SoccerAmericaMigrator.

- improved reusable command `cmd_set_first_image_from_content_as_featured_image` so that it runs on machines with less memory


iuravic:
Migrated and launched.

PR #291, opened 2023-05-10
add: Newsroom NZ migration script for subtitles

dkoo:
Adds a WP CLI script for `1200550061930446/1204502855407758`. Queries all posts that start with a bolded and italicized paragraph (NZ's standard format for subtitles) and moves those paragraphs to the `newspack_post_subtitle` meta field instead.

To test, ensure you have some posts whose first block is a paragraph containing bolded + italicized text. run `wp newspack-content-migrator newsroomconz-migrate-subtitles`. Supports a `--dry-run` flag to print changes without persisting.

Hint: a lot of whitespace changes from my IDE, so disable whitespace changes to see the true changeset.


kariae (approved):
LGTM

PR #311, opened 2023-06-28
CDiff recreate categories by name and parent term_id, skipping category description

iuravic:
### Change -- **CDiff now matches/searches/compares categories on Live VS categories on Staging by referencing their `name` and `parent term_id`, and stoped referencing category `description`**

Explanation -- when CDiff merges categories from live site to staging, it recreates the whole category tree from live on staging. It used to match/search a categories by its "name", "parent term_id" and "description".

But on some sites, category "description" gets changed on Live during migration, which broke category matching/searching back on Staging.

Now excluding category's "description" from category matching/searching, and category is identified by its "name" and "parent term_id".

In retrospect, checking for category "description" was an overkill and really unnecessary, buy now it's also causing issues. Even though we predicted that category changes wouldn't be made during migration, it happens sometimes since it's out of our control.

The new method `get_category_array_by_name_and_parent` is just a clone of existing `get_category_array_by_name_description_and_parent`, but without description. I left the other method untouched, just added this new one.

### DIFF
--- a/src/Logic/ContentDiffMigrator.php
+++ b/src/Logic/ContentDiffMigrator.php
@@ -614,7 +614,7 @@ class ContentDiffMigrator {
 		if ( 0 == $category_tree['parent'] ) {
 
 			// Get or create this top parent category.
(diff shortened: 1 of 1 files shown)


kariae (approved):
LGTM, thanks

PR #320, opened 2023-07-20
The Emancipator

iuravic:
The data for this migration is already in a WP database. All the stuff I run in this code is looking at the large serialized array in post_meta (with the key `api_content_element`).

Here is what goes on:
* We remove the categories "opinion" and "the-emancipator" and make sure the categories nested under them now has a parent of `0` instead of the deleted categories
* We add co-authors as the byline
* We grab descriptions of co-authors from the post content and delete it from post content after.
* We add post subtitles.

I also added a few functions to the logic in this repo. Most notably I added a generator function `get_all_wp_posts` that should be a little less agressive on memory than using `get_posts`.

## How to test 
(I've just been running on local)

```bash
# Empty the trash
wp post delete $( wp post list --post_status=trash --type=post_type --format=ids )

wp newspack-content-migrator emancipator-taxonomy
# wp newspack-content-migrator emancipator-authors
wp newspack-content-migrator emancipator-bylines
wp newspack-content-migrator emancipator-post-subtitles
# wp newspack-content-migrator emancipator-redirects
```
There are two commands commented out – those I still need some input on.

I'm not sure how much/little to log, so it's a bit all over the place :)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -26,7 +26,8 @@
     "symfony/css-selector": "^5.2",
     "automattic/newspack-cms-importers": "dev-master",
     "automattic/newspack-post-image-downloader": "dev-master",
-    "automattic/newspack-content-converter": "dev-master"
+    "automattic/newspack-content-converter": "dev-master",
+    "halaxa/json-machine": "^1.1"
   },
   "require-dev": {
     "automattic/vipwpcs": "^2.0",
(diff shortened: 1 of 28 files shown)


naxoc:
@iuravic I can't add you as the reviewer since you originally posted this one, but it's ready for review :)

naxoc:
This ran on a staging clone. I'll merge so future changes are easier to keep track of.

PR #322, opened 2023-07-21
The City, Chorus CMS migrator

iuravic:
Chorus CMS JSON export to WP importer.


iuravic:
@kariae merging now because I need these new GutenbergBlockGenerator methods for another Publisher.

If you have suggestions for improvements, please let me know and I'll refactor additionally.

PR #328, opened 2023-08-15
Retro Report: Update code after JSON api has changed

naxoc:
This PR makes the migrator for RR run again. 

* There were some changes to the API url, so some filename massage was needed
* The token is no longer needed, so I added a return rather than delete the function in case we need it again.
* I added a new helper class to read JSON with an iterator to save some memory.
* I cleaned up a little bit with some wrong return types and some confusing array juggling.

Note that I haven't changed anything in how the content is migrated.

Also note that I'm adding the JsonMachine library: https://github.com/halaxa/json-machine – I don't know that we read from all that much JSON, but it's a very handy and easy to use library that makes iterative reading much easier and lightweight.


src/Logic/JsonIterator.php
```diff
@@ -0,0 +1,63 @@
...
+	 *
+	 * @param string $filename Path to the JSON file – can be a URL too.
+	 * @param array $options Options to pass to JsonMachine.
+	 *
+	 * @return iterable
+	 */
+	public function items( string $filename, array $options = [] ): iterable {
+
+		// Suppress errors with '@' because we want to specifically errors below.
+		$stream = @fopen( $filename, 'r' );
```
eddiesshop, line 47: Do need to worry about closing the stream?

src/Logic/JsonIterator.php
```diff
@@ -0,0 +1,63 @@
...
+		if ( ! $stream ) {
+			$this->logger->log( self::LOG_NAME, "The JSON file with url {$filename} doesn't exist." );
```
eddiesshop, line 49: This is already awesome in and of itself, but could we change it to "The JSON resource at '{$filename}' doesn't exist." or even more simply "Doesn't exist: {$filename}"

eddiesshop (commented):
This is awesome! Really excited about using this package, especially for one of my current publishers I'm working on right now. Thanks for sharing this with us!

src/Logic/JsonIterator.php
naxoc: Done!

kariae (approved):

src/Logic/JsonIterator.php
kariae: That's the library I used for one of our publisher's JSON files (around 3GB) and it worked well, thanks for adding this one, LGTM.

PR #330, opened 2023-08-17
feat: logic func which will throw exception if core WP tables are empty.

eddiesshop:
During a recent content diff, somehow the live wp_terms table (the table with new content that we're trying to merge into the current database) was empty. This caused the function recreate_categories to not function as expected. This is because when it tries to gather all the categories which must be brought over, it does a JOIN with this wp_terms table. But if that table is empty, the query will return no results, even though the wp_terms_taxonomy table has valid categories to be imported.

This function attempts to prevent that from happening in the first place, by checking if any core WP tables are empty before proceeding in earnest with the content diff.

### DIFF
--- a/src/Command/General/ContentDiffMigrator.php
+++ b/src/Command/General/ContentDiffMigrator.php
@@ -1292,15 +1292,18 @@ class ContentDiffMigrator implements InterfaceCommand {
 	 * @param string $live_table_prefix Live table prefix.
 	 * @param array  $skip_tables       Core WP DB tables to skip (without prefix).
 	 *
-	 * @throws \RuntimeException In case that table collations do not match.
+	 * @throws \RuntimeException In case that table collations do not match, or any core WP tables are empty.
 	 *
 	 * @return void
 	 */
 	public function validate_db_tables( string $live_table_prefix, array $skip_tables ): void {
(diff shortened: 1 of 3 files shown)


eddiesshop:
@kariae during my local testing, I saw that the `wp_links` table was empty. I think we typically don't consider this a crucial table that must be migrated correct?

If that's the case, then we should probably add one more change to this PR which ignores that table by default. Let me know what you think though.

iuravic (changes requested):
@eddiesshop as I shared during our Sprint, I think we should hardcode that some core tables are expected to possibly be empty not to break automatic execution of content refresh via Knife.

I've created a blank Atomic site and the following tables can be empty:
`wp_commentmeta wp_comments wp_links wp_termmeta`

Can you please add this?

eddiesshop:
@iuravic I added those tables you indicated to the ignore list. So if an exception is thrown, it is caught, and if those tables are the reason for the exception, then they are ignored, and the content diff is allowed to continue.

eddiesshop:
Hey @iuravic! I ran PHPCS on the code that was modified. I've tested this fix with the content refresh that I was performing when this problem arose. Is there a particular site/setup you'd like me to test this with?

kariae (approved):
Seems like all the requests from the review were, I think we can merge this.

PR #331, opened 2023-08-24
feat: add command to link all users to GAs

leogermani:
Adds a new command to the CoAuthorsPlus migrator class

It will loop through all WP users and look for a Guest Author with the same email.

If the guest author is not already linked to an account, it will link the guest author to that account.

Tested on NewsroomNZ already

### DIFF
--- a/src/Command/General/CoAuthorPlusMigrator.php
+++ b/src/Command/General/CoAuthorPlusMigrator.php
@@ -7,6 +7,7 @@ use \NewspackCustomContentMigrator\Command\InterfaceCommand;
 use \NewspackCustomContentMigrator\Logic\CoAuthorPlus;
 use \NewspackCustomContentMigrator\Logic\Posts;
 use \NewspackCustomContentMigrator\PluginSetup;
+use \NewspackCustomContentMigrator\Utils\Logger;
 use \NewspackCustomContentMigrator\Utils\PHP;
 use \WP_CLI;
 use \WP_Query;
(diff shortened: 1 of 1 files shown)


src/Command/General/CoAuthorPlusMigrator.php
```diff
@@ -1331,4 +1353,66 @@ private function get_posts( $post_type ) {
...
+
+	}
+
+	/**
+	 * Simple file logging.
+	 *
+	 * @param string $file    File name or path.
+	 * @param string $message Log message.
+	 */
+	public function log( $file, $message ) {
```
naxoc, line 1413: It looks like the Logger instance is used instead of this one. Can it be deleted?

src/Command/General/CoAuthorPlusMigrator.php
```diff
@@ -1331,4 +1353,66 @@ private function get_posts( $post_type ) {
...
+		} else {
+			WP_CLI::line( 'This command will modify the database.');
+			WP_CLI::line( 'Consider running it with --dry-run first to see what it will do.');
+			WP_CLI::confirm( "Are you sure you want to continue?", $assoc_args );
+		}
+
+		$guest_authors = new \CoAuthors_Guest_Authors();
+
+		// Get the list of users.
+		$users = get_users();
```
naxoc, line 1378: I could worry that on a large site with many users this would be a heavy query. If we only need the ID – we could maybe use a raw query for them instead?

naxoc (commented):
I added a couple of small stuff comments, but this is a great addition. Thanks!

src/Command/General/CoAuthorPlusMigrator.php
leogermani: We actually only need the user_email, but `link_guest_author_to_wp_user` takes the WP_User object as a parameter

We could just add the post_meta ourselves... or we could just get the user if needed to pass to that function.

However, if I refactored that now I would not be able to test it in a good way... I think I rather stick with the tested code for now and address this if it becomes an issue in a future migration

PR #332, opened 2023-08-25
LSV: Fix user slugs

naxoc:
I'm not confident this will not cause other problems, but this fixes the slugs for authors.

It's pretty heavy handed, but seems to get the job done. There are two commands - one for users and one for guest authors. Run the user one first.

```shell
wp newspack-content-migrator la-silla-vacia-update-user-slugs
wp newspack-content-migrator la-silla-vacia-update-guest-author-slugs
```

There are a lot of users to fix now. On my local machine the user slug update takes ~15 minutes to run. 

Some users have no first name, last name, or anything at all to make a slug from. I skip them because I don't know what to make a slug from.

### DIFF
--- a/src/Command/PublisherSpecific/LaSillaVaciaMigrator.php
+++ b/src/Command/PublisherSpecific/LaSillaVaciaMigrator.php
@@ -2546,16 +2546,17 @@ BLOCK;
 				update_post_meta( $post_id, 'cap-user_login', $slug );
 
 				$sql    = $wpdb->prepare(
-					"UPDATE {$wpdb->terms} SET slug = '%s' WHERE slug = '%s'",
-					$slug,
-					'cap-' . ( $data['old_nicename'] ?? $old_post_name )
+					"UPDATE {$wpdb->terms} SET slug = '%s', name = '%s' WHERE slug = '%s'",
+					$post->post_name,
+					$post->post_title,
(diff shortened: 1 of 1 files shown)


src/Command/PublisherSpecific/LaSillaVaciaMigrator.php
```diff
@@ -2409,6 +2426,106 @@ public function update_podcasts( array $args, array $assoc_args) : void {
...
+				}
+				update_post_meta( $post_id, 'cap-user_login', $slug );
+
+				$sql    = $wpdb->prepare(
+					"UPDATE {$wpdb->terms} SET slug = '%s' WHERE slug = '%s'",
+					$post->post_name,
+					'cap-' . ( $data['old_nicename'] ?? $old_post_name )
+				);
+				$result = $wpdb->query( $sql );
+
```
eddiesshop, line 2519: Have you had a chance to investigate whether `cap-linked_account` needs to be updated also? That seems to be tied to `user_login`.

naxoc: Just checked – there are 9 :)

naxoc:
OK, I looked into the linked account thing and it still works. I think this one is ready to run.

naxoc:
Closing this one as it's been resolved in other ways

PR #336, opened 2023-08-31
Try to guess the file extension for external file downloads without one

naxoc:
This fixes an edge case in `import_external_file()` where files would not get imported if the url (or file path) they are on does not have a file extension. It uses the mime type to try to sniff out what kind of file it could be. Not 100% good but a lot better than importing 0%


iuravic (approved):
Awesome! Thanks!

PR #337, opened 2023-09-01
Try to find an existing file in the ML when adding external files

naxoc:
This adds a function that checks if we already have an upload to the media library of a given file.
The file to check needs to be in the filesystem (like we don with tmp files in `import_external_file()` in `Attachments.php`).

The idea is to check if we already have a file uploaded by getting file(s) from the media library with the same filename and then checking to see if a file has:
* The same file size 
* The same hash of the file content

We talked about this one in Slack: p1693569468634049-slack-C013N9S2C7Q

### DIFF
--- a/src/Logic/Attachments.php
+++ b/src/Logic/Attachments.php
@@ -106,12 +106,14 @@ class Attachments {
 	}
 
 	/**
-	 * @param string $filepath
-	 * @param string $filename
+	 * Try to get the attachment ID for a file if one just like it has already been uploaded.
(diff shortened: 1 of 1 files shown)


iuravic:
😮 😮 😮  ... I tested it and it works GREAT!
Awesome possum! 🥇🥇🥇
Thank you 🙇

eddiesshop (approved):
Nice! Like the thought about checking file sizes first before comparing hashes! Very cool!

PR #338, opened 2023-09-04
Add filter logic to change upload destination path

naxoc:
When importing articles/issues/etc. that have a publish date, this allows us to set the upload destination dir to match the date for instance.

I needed the instance variable to make this work, but if there is a more elegant way we should remove that :) 

To test it and use it – call `add_upload_path_filter( $pub_date )` adding uploads and remove again with `remove_upload_path_filter()`. This can be done in an import loop for instance.

### DIFF
--- a/src/Logic/Attachments.php
+++ b/src/Logic/Attachments.php
@@ -5,10 +5,15 @@ namespace NewspackCustomContentMigrator\Logic;
 use \WP_CLI;
 
 class Attachments {
+
+	/**
+	 * @var null|string The subdirectory to use for the upload directory when the add_upload_path_filter is active.
+	 */
+	private $upload_subdir = null;
+
+
 	/**
 	 * Imports a media object from file and returns the ID.
-	 *
-	 * @deprecated Warning: this function is quite slow, and the new function `import_external_file` here should be used instead.
 	 *
 	 * @param string $file Media file full path.
 	 *
@@ -101,7 +106,7 @@ class Attachments {
 	 * @param integer   $posts_per_batch Total of posts tohandle per batch.
 	 * @param integer   $batch Current batch in the loop.
 	 * @param integer   $start_index Index from where to start the loop.
-	 * @param func|null $logger Method to log results.
+	 * @param callable|null $logger Method to log results.
 	 *
 	 * @return mixed[] Array of the broken URLs indexed by the post IDs.
 	 */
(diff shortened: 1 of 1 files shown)


src/Logic/Attachments.php
```diff
@@ -5,11 +5,16 @@
 use \WP_CLI;
 
 class Attachments {
```
iuravic, line 7: For all reusable code, we should also run PHPCS 👍 Let's please do that before PR-ing.

Here's how to do it:
1. `phpcbf` will first fix all the CS things automatically which it can, `vendor/bin/phpcbf phpcs.xml src/Logic/Attachments.php`
2. `phpcs` will then tell you which `ERRORs` still need to be fixed manually (warnings can be disregarded), `vendor/bin/phpcs phpcs.xml src/Logic/Attachments.php`

Looks like someone introduced some errors before this commit. It's not your responsibility to fix those preexisting errors, but if you're able to (or even some of them), it will be appreciated.

src/Logic/Attachments.php
```diff
@@ -257,4 +262,42 @@ public function get_attachment_by_filename( $filename ) {
 
 		return $attachment_id;
 	}
+
+	/**
+	 * Change the subdirectory in the uploads folder that files go to.
+	 *
+	 * Don't forget to remove the filter again after you're done – use remove_upload_path_filter().
+	 *
+	 * @param string $path Subdirectory you want uploads to go to.
```
iuravic, line 271: Can we give an example string here? I think it would help a lot -- e.g. do we use the full path?, just the part from uploads/..., or something different?

iuravic:
I love the concept and purpose of this! But if I can just ask/double-check how it's actually supposed to be used, and perhaps explore alternative routes?

I hope I understood correctly that the point of running `add_upload_path_filter()` is to set the subfolder path, so that `import_external_file()` would then save it to a desired subfolder, not just any subfolder (like current date)?

If that's the case, let me also point out that by setting the `$post_id` argument for `import_external_file()` should also get it to be uploaded to the same date-subfolder of the Post ID's date. `$post_id` being the Post object to which this attachment "belongs to", so it gets saved to same date-subfolder as the Post's date.

But first back to checking whether I understood how to use this these functions. Are these steps correct?:
1. first initiate the object from class
2. second, run `add_upload_path_filter()`
3. then after the public property `$this->upload_subdir` has been set, we can now run `import_external_file()` <-- is this how it's supposed to be used?
3. and if I got these correct, how is `remove_upload_path_filter()` used?

I would propose is that we explore if we can avoid setting a value to an object property which then becomes a condition to how a method will work. Two reasons:
- this now becomes a [state object/pattern](https://en.wikipedia.org/wiki/State_pattern), and we must actually keep track of how a method is executed, not just run it
- I would propose that we perhaps add just a new optional method argument (`string $path`) and set the filter during method execution if the argument was provided. This seems simpler to me -- though I'm not a 100% sure I understood how things are supposed to be run above ☝️ 
- though my first vote would be to avoid creating a state object, if we still decide to have it, perhaps we would want to also have the `construct`or get `string $path` and set it as the internal property, while keeping both the setter and getter methods
- and lastly, if we keep them, I'd recommend renaming add_ to set_ method name, just to make it a bit more "getter"-"setter" standardized, though this is just a very tiny remark, not important at all.

iuravic (changes requested):
Added a couple of remarks and questions.

naxoc:
> If that's the case, let me also point out that by setting the $post_id argument for import_external_file() should also get the attacment to be uploaded to the same date-subfolder as the Post ID's date

We have code where the file gets created before the post, so the ID would not be available. I suppose we could re-write some of that code to wait for a post ID. That might be easier. I don't love the instance variable either.

> But first back to checking whether I understood how to use this these functions. Are these steps correct?:
>
> first initiate the object from class

I've mostly seen the class instanciated in an importer's constructor, but it could also be done directly in a function.

> second, run add_upload_path_filter()

Sorry if it was unclear. I would use this in a loop, so that is why the filter can be added and removed. So something like this
```php
foreach ($json_line as $obj) {
 add_upload_path_filter( $obj->pub_date );
$attachment_id = $this->attachments->import_external_file( $obj->file_path );
remove_upload_path_filter();
}
```
The date can vary for each run of the loop.

> then after the public property $this->upload_subdir has been set, we can now run import_external_file() <-- is this how it's supposed to be used?
Well, kinda. The filter is added and removed pr. import file.

> and if I got these correct, how is remove_upload_path_filter() used? Perhaps just as a convenience function to remove the "special subfolder" that's being set for future attachment imports by the same object?

Yes, that should be called every time a file has been uploaded.

The `$upload_subdir` is relative to the uploads folder. E.g. `2023/05`.

All that said – maybe it makes more sense to make sure that a post exists before importing a file. Let me think about that a bit :)

naxoc:
I'll close this for now. It can be a bit of work to create a post before any attachments because of validation of files etc, but I think it's better than adding and removing a filter. Thanks for the input!

PR #340, opened 2023-09-11
Download missing images

naxoc:
This downloads missing images and is probably re-usable for other sites in the future.

I pulled most code into a new file because there ended up being a lot of it :)

Run it with `wp newspack-content-migrator la-silla-vacia-download-missing-images --media-dir="../path/to/medi" --post-id=5325` and post id is optional.

### DIFF
--- a/src/Command/General/DownloadMissingImages.php
+++ b/src/Command/General/DownloadMissingImages.php
@@ -1,0 +1,345 @@
+<?php
+
+namespace NewspackCustomContentMigrator\Command\General;
+
+use \NewspackCustomContentMigrator\Command\InterfaceCommand;
+use NewspackCustomContentMigrator\Logic\Attachments;
+use NewspackCustomContentMigrator\Logic\Posts;
+use NewspackCustomContentMigrator\Utils\Logger;
+use NewspackCustomContentMigrator\Utils\MigrationMeta;
+use WP_CLI;
+
+class DownloadMissingImages implements InterfaceCommand {
+
+	/**
+	 * @var null|self
+	 */
+	private static $instance = null;
+
+	private $command_meta_key = 'download_missing_images';
+	private $command_meta_version;
+	private $log_file;
+
+	/**
+	 * @var Attachments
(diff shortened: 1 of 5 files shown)


src/Command/General/DownloadMissingImages.php
```diff
@@ -0,0 +1,345 @@
...
+			}
+		}
+
+		return false;
+	}
+
+	public function move_local_file( string $from, string $to ): bool {
+		// Make sure the destination folder exists.
+		wp_mkdir_p( dirname( $to ) );
+
```
eddiesshop, line 307: Can we please log any files that are moved?

src/Command/General/DownloadMissingImages.php
```diff
@@ -0,0 +1,345 @@
...
+				wp_update_post( $data['post'] );
+			}
+
+			if ( ! empty( $urls_not_found ) ) {
+				$this->logger->log( $this->log_file,
+					sprintf( "Could not find images for these images for post with ID %d: %s\n\t", $data['post']->ID, implode( "\n\t", $urls_not_found ) ),
+					Logger::WARNING );
+			}
+
+			MigrationMeta::update( $data['post']->ID, $this->command_meta_key, 'post', $this->command_meta_version ); // TODO
```
eddiesshop, line 170: Not a big deal, but is this TODO still relevant?

eddiesshop (changes requested):
Super clean and efficient! Just 2 very small things and I think we're good to go!

src/Command/General/DownloadMissingImages.php
naxoc: Yup! I added that

naxoc:
I added one more arg to be used like this: `--post-id-range=1000-20000"` so we can run post id ranges.

eddiesshop:
Tested this out on a few posts and it's working great! Thank you for all this hard work!

eddiesshop (approved):

PR #342, opened 2023-09-14
Mustang News: Command to move image credit from posts to attachments

naxoc:
- [x] We need to clean up the class inclusion in newspack-custom-content-migrator.php once the command has been run.


naxoc:
This was run a while ago, so I'm merging this one.

PR #344, opened 2023-09-18
Do not always install WP importer

iuravic:
Super simple PR -- removes one line from main plugin file.

WP Importer should not always be installed and activated every time this plugin is used.


naxoc (approved):
Looks good!

PR #350, opened 2023-09-27
Remove PHP unions for PHP 7 compat

naxoc:
## How to test
No testing can be done, but look at the code :) 

---

### DIFF
--- a/src/Logic/Taxonomy.php
+++ b/src/Logic/Taxonomy.php
@@ -90,7 +90,7 @@ class Taxonomy {
 	 *
 	 * @return string|null Term ID or null if not found.
 	 */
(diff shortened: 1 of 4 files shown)


src/Logic/Taxonomy.php
```diff
@@ -118,7 +118,7 @@ public function get_term_id_by_taxonmy_name_and_parent( string $taxonomy, string
 	 *
 	 * @return string Category term ID.
```
iuravic, line 119: Could we also update this docblock to:
`* @return string|null Category term ID.`

iuravic (changes requested):

PR #351, opened 2023-09-28
Load publishers conditionally based on PHP version

naxoc:
This is a small attempt to make it possible for us to use PHP8 on publishers while still keeping PHP7 compatibility in the plugin. I added a conditional where PHP8 publishers can be added.

## How to test
* Run `wp newspack-content-migrator emancipator-list-content-refresh` (it does nothing other than list commands, so it's safe). If it loads correctly (and your env is on PHP 8) we are good. For extra points, also test on a PHP 7  site and verify that the command would not work.

---


naxoc:
I'm closing this after we talked about it on a call. We have other tricks up our sleves :)

PR #353, opened 2023-09-29
Fix the "Sorry, you are not allowed to upload this file type." error

naxoc:
Also add a wrapper for file import that enforces a post id to attach it to. This should keep a neater organization into year folders in upload.

I also added one more extra check so we get fewer file not found errors.

It helps to import a file which has a different extension (e.g. .blob instead of .jpeg, missing .pdf, etc).

## How to test
- try calling both `import_external_file()` and `import_attachment_for_post()` including the `$desired_filename` argument on a file that you've changed the file extension to something like `.notallowed` and verify that you get the file downloaded correctly.

---

### DIFF
--- a/src/Logic/Attachments.php
+++ b/src/Logic/Attachments.php
@@ -23,6 +23,24 @@ class Attachments {
 	}
 
 	/**
+	 * Wrapper for import_external_file() with fewer args and enforced post_id.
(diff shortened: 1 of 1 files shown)


kariae (approved):
LGTM, thanks

PR #355, opened 2023-10-03
Fix/moving content from a category to another

kariae:
## How to test
- Create two categories "Source Category", and "Destination Category".
- Create Three posts, one for each category, and the third for both categories.
- Run the command to move posts from the "Source Category" to the "Destination Category": `wp newspack-content-migrator move-content-from-one-category-to-another --source-term-id=123 --destination-term-id=456`
- Notice the SQL error and that the posts were not moved.
- Switch to this branch and re-run the command.
- Notice The Source Category with no posts, and all the posts were moved to the Destination Category.

Play a bit more with it 😅 

---

### DIFF
--- a/src/Command/General/TaxonomyMigrator.php
+++ b/src/Command/General/TaxonomyMigrator.php
@@ -379,10 +379,10 @@ class TaxonomyMigrator implements InterfaceCommand {
 		// Check IDs.
 		$source_term      = get_term( $source_term_id, $taxonomy );
 		$destination_term = get_term( $destination_term_id, $taxonomy );
-		if ( is_null( $source_term ) ) {
+		if ( ! $source_term instanceof \WP_Term ) {
 			WP_CLI::error( 'Wrong source term ID.' );
 		}
-		if ( is_null( $destination_term ) ) {
+		if ( ! $destination_term instanceof \WP_Term ) {
 			WP_CLI::error( 'Wrong destination term ID.' );
 		}
 		if ( $source_term_id == $destination_term_id ) {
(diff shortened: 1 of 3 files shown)


src/Command/General/TaxonomyMigrator.php
```diff
@@ -357,38 +364,38 @@ public function cmd_move_category_tree( $pos_args, $assoc_args ) {
...
+		$taxonomy            = $assoc_args['taxonomy'] ?? 'category';
 
 		// Check IDs.
-		$source_category      = get_category( $source_term_id );
-		$destination_category = get_category( $destination_term_id );
-		if ( is_null( $source_category ) ) {
-			WP_CLI::error( 'Wrong source category ID.' );
+		$source_term      = get_term( $source_term_id, $taxonomy );
+		$destination_term = get_term( $destination_term_id, $taxonomy );
+		if ( is_null( $source_term ) ) {
```
naxoc, line 382: `get_term()` is one of those lovely WP functions that can return many different types. I wonder if it would be better to check if the terms are an instance of WP_Term instead? `null` would only be on failure – not failure to find the term.

naxoc (commented):
I tested this and it worked great.  I just had a small comment :)

Here are the commands I used:

```bash
❯ wp term create category  "Source Cat"
Success: Created category 1327.
❯ wp term create category  "Destination Cat"
Success: Created category 1328.

wp post create --post_type=post --post_title='Source post' --post_status=publish --post_category=1327

wp post create --post_type=post --post_title='Dest post' --post_status=publish --post_category=1328

wp post create --post_type=post --post_title='Both post' --post_status=publish --post_category=1327,1328

# Run old command (nothing happened)
wp newspack-content-migrator move-content-from-one-category-to-another --source-term-id=1327 --destination-term-id=1328

# Run new command
wp newspack-content-migrator move-content-from-one-term-to-another --taxonomy=category --source-term-id=1327 --destination-term-id=1328 

```

src/Command/General/TaxonomyMigrator.php
kariae: Nice catch @naxoc fixed, can you do a final review, please?

naxoc (approved):
Looks good! 
 OK

PR #357, opened 2023-10-04
Embarcadero Migrator, import comments wo/ user, and smaller updates

iuravic:
- Command `embarcadero-migrate-comments` now imports comments even if coment's user is not found (WP accepts comments without a WP_User attached to them)

I also added a helper command with a workflow which helped me fix a broken CSV for Pleasanton. It's a bit elaborate/specific, but it does the trick. I tried several variations, and this one worked:
- Added the helper command `embarcadero-helper-fix-tsv-file` which helps fix broken CSVs. I've shared how it's used up in the comments at the top of the class -- see "How to fix CSVs if migrator reports that some rows can't be read":
    1. convert CSV to TSV using OSX Pages. I first converted CSV to TSV this because commas are a part of the problem, and tab-separated-values bypass all that. Also everyone from our Team has Pages and it works well (even with 300MB CSVs)
    2. feed the the resulting TSV through my helper command as argument -- iterates through all records, and fixes the  incompatible `\""` or `\"` escapings of the the double quote
    3. then use the resulting fixed TSV instead of CSV for the importer command (in my case it was the import comments and the broken comments.csv)
- Updated existing function `get_data_from_csv` to `get_data_from_csv_or_tsv` so that it now works with both CSVs and TSVs

## How to test
- this is an edge case -- in case of a broken CSV, follow instructions under "One method to fix CSVs which can't be read" to get a fixed TSV and use it instead of the original CSV
- if a record in comments.csv references an user which is not in the users CSV, it will still get imported with blank user data

---

### DIFF
--- a/src/Command/General/EmbarcaderoMigrator.php
+++ b/src/Command/General/EmbarcaderoMigrator.php
@@ -40,6 +40,7 @@ use \WP_CLI;
  *     - wp newspack-content-migrator embarcadero-migrate-comments \
  *       --comments-csv-file-path=/path/to/comments.csv \
  *       --comments-zones-file-path=/path/to/comment_zones.csv
+ *       --users-file-path==/path/to/registrated_users.csv
  *
  *     - wp newspack-content-migrator embarcadero-migrate-print-issues \
  *       --publication-name="Danville San Ramon" \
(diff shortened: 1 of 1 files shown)


kariae (approved):

kariae:
Thanks @iuravic, I'm merging this!

PR #374, opened 2023-11-03
Town News: Try to avoid using email as display name for authors

naxoc:
I noticed that we can risk creating slugs with emails. This does some string juggling to avoid that.

## How to test
- Run the migration and verify that guest users are created with whatever is before the '@' if there is only an email on them.


src/Command/General/TownNewsMigrator.php
```diff
@@ -696,7 +698,7 @@ private function set_post_author( $post_id, $display_name, $author_meta, $defaul
 			);
 		} else {
 			// Set as a co-author.
-			$guest_author = $this->coauthorsplus_logic->get_guest_author_by_user_login( $email );
+			$guest_author = $this->coauthorsplus_logic->get_guest_author_by_user_login( $display_name );
```
kariae, line 701: Using `$email` was a bug here, nice catch! The [create_guest_author](https://github.com/Automattic/newspack-custom-content-migrator/blob/ea5e7a66d1c0727230936284059fbd9e68657519/src/Logic/CoAuthorPlus.php#L105) used [here](https://github.com/Automattic/newspack-custom-content-migrator/pull/374/files#diff-a630c53224b300be80052c56980d3a5f4a5d0ed96e9764671318e1fa9010aaa7R714) to create a new co-author uses the sanitized display_name as user_login in case we don't provide the later, so I think to fetch co-authors using `$display_name` as user_login we should pass the sanitized version.
```suggestion
			$guest_author = $this->coauthorsplus_logic->get_guest_author_by_user_login( sanitize_title( $display_name ) );
```

kariae (changes requested):

src/Command/General/TownNewsMigrator.php
naxoc: Thanks Zak - good catch! 

I added sanitization with sanitize_user because I assume that is what the user login is run through when it's created. Does that make sense?

naxoc:
I also added tag stripping in bylines – I had a couple of files where the byline was dirty.

src/Command/General/TownNewsMigrator.php
kariae: The user is created as of now with [`sanitize_title`](https://github.com/Automattic/newspack-custom-content-migrator/blob/ea5e7a66d1c0727230936284059fbd9e68657519/src/Logic/CoAuthorPlus.php#L112), but I think it's better to fix it properly and change it there to `sanitize_user`

naxoc:
@kariae This one got a bit old. I'm not sure if it is still relevant. What do you think?

naxoc:
@kariae I think this one can just be closed with no merge. It's so old :) What do you think?

kariae (approved):
LGTM!

PR #375, opened 2023-11-03
TN: Add a primary category

naxoc:
Those are handy for the url structure.

## How to test
- Run the content migrator and verify in the GB sidebar that posts have a primary category.

### DIFF
--- a/src/Command/General/TownNewsMigrator.php
+++ b/src/Command/General/TownNewsMigrator.php
@@ -483,7 +483,6 @@ class TownNewsMigrator implements InterfaceCommand {
 
 			// Create the categories hierarchy.
 			foreach ( $category_hierarchy as $category_index => $category ) {
-				$category_id   = 0;
 				$category_name = $this->get_category_name_from_tn_slug( $category );
 				if ( 0 !== $category_index ) {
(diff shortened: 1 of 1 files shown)


src/Command/General/TownNewsMigrator.php
```diff
@@ -483,7 +483,6 @@ function( $keyword ) {
 
 			// Create the categories hierarchy.
 			foreach ( $category_hierarchy as $category_index => $category ) {
-				$category_id   = 0;
```
kariae, line 486: Nice catch for the primary category, Camilla. We need to initialize $category_id here to 0 or change the way we check the existence of it in line 495, otherwise, this looks good to me, thanks.

kariae (changes requested):

src/Command/General/TownNewsMigrator.php
naxoc: It would always be set because of the `else` but I realized that the check in line 494 is checking for a number – not an error, so I added a check. Does that make sense?

kariae: Yup, perfecto!

kariae (approved):

PR #379, opened 2023-11-10
fix(EmbarcaderoMigrator): set authors as WP users when possible

kariae:
## How to test
Run the migration command and see the number of contributors in the users list increment.

### DIFF
--- a/src/Command/General/EmbarcaderoMigrator.php
+++ b/src/Command/General/EmbarcaderoMigrator.php
@@ -537,7 +537,7 @@ class EmbarcaderoMigrator implements InterfaceCommand {
 			$this->logger->log( self::LOG_FILE, sprintf( 'Importing post %d/%d: %d', $post_index + 1, count( $posts ), $post['story_id'] ), Logger::LINE );
 
 			$wp_contributor_id = null;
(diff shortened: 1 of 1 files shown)


src/Command/General/EmbarcaderoMigrator.php
```diff
@@ -519,15 +536,10 @@ function( $post ) use ( $imported_original_ids ) {
...
 			$this->logger->log( self::LOG_FILE, sprintf( 'Importing post %d/%d: %d', $post_index + 1, count( $posts ), $post['story_id'] ), Logger::LINE );
 
-			$contributor_index = array_search( $post['byline'], array_column( $contributors, 'full_name' ) );
-
 			$wp_contributor_id = null;
-			if ( false !== $contributor_index ) {
-				$contributor       = $contributors[ $contributor_index ];
-				$wp_contributor_id = $this->get_or_create_user( $contributor['full_name'], $contributor['email_address'], 'contributor' );
-
+			if ( isset( $post['author_email'] ) && ! empty( $post['byline'] ) && ! empty( $post['author_email'] ) ) {
```
naxoc, line 540: Small thing: No need to check `$post['author_email']` twice. The empty() call also checks if it is set.

kariae: Awesome, thanks, Camilla!

PR #381, opened 2023-11-24
fix(EmbarcaderoMigrator): use the correct topic ID when importing comments

kariae:
Import stories missing `topic_id`, and use it to correctly import comments.

## How to test
- Re-run the `embarcadero-import-posts-content` with the flags `--refresh-content --skip-post-content`
- Clean all imported comments from the database (purge wp_comments and wp_commentmeta tables).
- Re-run `embarcadero-migrate-comments` command.
- Check a few posts and make sure the comments are imported to the correct posts.

---


naxoc (approved):
Looks good! 

One small nit that you are free to ignore: If we check for empty instead of `0` we should catch more (probably unlikely) cases where the topic_id is not set correctly.

PR #387, opened 2023-12-11
MinnPost migrator

iuravic:


iuravic:
Moved to archive in separate https://github.com/Automattic/newspack-custom-content-migrator/pull/528

PR #391, opened 2023-12-22
Add re-usable class for blocks manipulation

naxoc:
This adds a class that can remove/replace/search/filter etc GB blocks. It only has what I needed for a specific migration, and we can certainly add more as we go.

## How to test
- The functions should do what they say – you will likely need to write some test code to run them - or look at the code :) 

---


iuravic (approved):
🥇 👍 !

PR #393, opened 2023-12-29
fix(Attachment): sanize filename before search query

kariae:
## How to test
- Upload an image with a special character in the filename (e.g. `1*test.jpeg`) using the `import_external_file` method two times and notice that the image is duplicated in the media library.
- Pull this and redo the previous  step and notice that the media is not duplicated this time.

---

### DIFF
--- a/src/Logic/Attachments.php
+++ b/src/Logic/Attachments.php
@@ -25,11 +25,11 @@ class Attachments {
 	/**
 	 * Wrapper for import_external_file() with fewer args and enforced post_id.
 	 *
-	 * @param int $post_id              Post ID of the post the media should be attached to.
+	 * @param int    $post_id              Post ID of the post the media should be attached to.
 	 * @param string $path              Media file full URL or full local path, or URL to the media file.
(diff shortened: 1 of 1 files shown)


src/Logic/Attachments.php
```diff
@@ -149,7 +149,7 @@ public function maybe_get_existing_attachment_id( string $filepath, string $file
 		}
 
 		global $wpdb;
-		$like = '%' . $wpdb->esc_like( $filename );
+		$like = '%' . $wpdb->esc_like( sanitize_title( $filename ) );
```
naxoc, line 152: I think `sanitize_file_name()` is used for files on upload. Probably not a lot of difference, but maybe it is better to use that so there is a higer probability of a match?

naxoc (changes requested):

src/Logic/Attachments.php
kariae: nice catch!

naxoc (approved):
Beautiful :)  
 OK

PR #394, opened 2023-12-29
feat(CoAuthors): allow disabling CAP cache when getting linked authors

kariae:
## How to test
- No test needed, this add a parameter that already exists in the used `get_guest_author_by` from the CAP plugin.

---


naxoc (approved):
Good one!
 OK

PR #395, opened 2023-12-29
fix(GutenbergBlockGenerator): make sure the attahcment ID received is a number in get_image

kariae:
Retreived posts by $wpdb get their IDs as strings, if we pass this to get_image it generates a broken image block, this makes sure the ID is a number.

---


naxoc (approved):
Looks good! 
 OK

PR #396, opened 2023-12-29
feat(MediumMigrator): add Medium Migrator

kariae:
## How to test

---

### DIFF
--- a/src/Command/General/MediumMigrator.php
+++ b/src/Command/General/MediumMigrator.php
@@ -1,0 +1,321 @@
+<?php
+
+namespace NewspackCustomContentMigrator\Command\General;
+
+use \NewspackCustomContentMigrator\Command\InterfaceCommand;
+use \NewspackCustomContentMigrator\Logic\Medium;
+use NewspackCustomContentMigrator\Logic\Attachments;
+use NewspackCustomContentMigrator\Logic\SimpleLocalAvatars;
+use NewspackCustomContentMigrator\Utils\Logger;
+use Symfony\Component\DomCrawler\Crawler;
+use \WP_CLI;
+
+/**
+ * Migrates Medium archive.
+ */
+class MediumMigrator implements InterfaceCommand {
+	const ORIGINAL_ID_META_KEY = '_medium_original_id';
+
+	/**
+	 * Filename of where to save the log.
+	 *
+	 * @var string $log_file Log file name.
+	 */
+	private static $log_file = 'medium-migrator.log';
+
+	/**
+	 * Migrator instance.
+	 *
+	 * @var null|InterfaceCommand Instance.
+	 */
+	private static $instance = null;
+
+	/**
+	 * Medium logic instance.
+	 *
+	 * @var Medium
+	 */
+	private $medium_logic = null;
+
+	/**
+	 * Instance of Attachments Login
+	 *
+	 * @var null|Attachments
+	 */
+	private $attachments;
+
+	/**
+	 * Instance of SimpleLocalAvatars.
+	 *
+	 * @var null|SimpleLocalAvatars
+	 */
+	private $simple_local_avatars_logic;
+
+	/**
+	 * Logger instance.
+	 *
+	 * @var Logger.
+	 */
+	private $logger;
+
+	/**
+	 * Constructor.
+	 */
+	private function __construct() {
+		$this->medium_logic               = new Medium();
+		$this->attachments                = new Attachments();
+		$this->simple_local_avatars_logic = new SimpleLocalAvatars();
+		$this->logger                     = new Logger();
+	}
+
+	/**
+	 * Singleton get_instance().
+	 *
+	 * @return InterfaceCommand|null
+	 */
+	public static function get_instance() {
+		$class = get_called_class();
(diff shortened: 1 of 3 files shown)


src/Command/General/MediumMigrator.php
```diff
@@ -0,0 +1,333 @@
...
+	 * @param array $author Author data.
+	 *
+	 * @return int|bool Author ID or false on error.
+	 */
+	private function get_or_insert_author( $author ) {
+		$author_id = username_exists( $author['user_login'] );
+		if ( ! $author_id ) {
+			$author_id = wp_insert_user(
+				[
+					'user_login'   => $author['user_login'],
```
iuravic, line 168: I assume this won't cause any leaking emails, because we provided the user_login -- but pinging in @naxoc and @eddiesshop just to reassure us, because they have more EXP than me w/ this.

src/Command/General/MediumMigrator.php
```diff
@@ -0,0 +1,333 @@
...
+
+	/**
+	 * Import images from the Post content.
+	 *
+	 * @param string $post_content Post content.
+	 * @param int    $post_id      Post ID.
+	 *
+	 * @return string Post content with updated image URIs.
+	 */
+	private function import_post_images( $post_content, $post_id ) {
```
iuravic, line 204: This is certainly good. Just sharing that we also could have used https://github.com/Automattic/newspack-post-image-downloader as final step, which can downloads all externally hosted images (import attachments and update post_content URLs); It does a very similar thing already [here](https://github.com/Automattic/newspack-post-image-downloader/blob/master/src/class-downloader.php#L319). And it also handles both relative and absolute URLs, has ability to white- or blacklist hosts, etc.

But the plus side of your approach is that `import_external_file()` used here will prevent duplicate attachments, while the Newspack Post Image Downloader uses it's own version of "import_external_file()"which is not as good as our standard one, and will create dupes.

iuravic (approved):
Ran the import on my local and it worked beautifully!

We do have some:
- WARNING:  -- Error importing featured image: Sorry, you are not allowed to upload this file type.
Perhaps we could give a list of post IDs to the TAM / Publisher just to let them know how many were not allowed. Then maybe they can do those manually, because it seems there were just a few.

src/Command/General/MediumMigrator.php
naxoc: We should always supply `user_nicename` to make sure the slug does not become an email. It looks like we could use the `author['display_name']` value here for that?

src/Command/General/MediumMigrator.php
```diff
@@ -0,0 +1,333 @@
...
+		if ( is_wp_error( $post_id ) ) {
+			$this->logger->log( self::$log_file, 'Error: ' . $post_id->get_error_message(), Logger::LINE );
+			return;
+		}
+
+		$this->logger->log( self::$log_file, ' -- Article inserted with ID: ' . $post_id, Logger::LINE );
+
+		// Set the featured image.
+		if ( ! empty( $article['featured_image'] ) ) {
+			$featured_image_id = $this->attachments->import_external_file( $article['featured_image']['url'], $article['title'], $article['featured_image']['caption'] );
```
naxoc, line 302: Not a blocker at all, but I wanted to mention that using `import_attachment_for_post()` in the Attachment class enforces passing the post id. It's not important, but it makes for a neater organization by date in the uploads folder.

src/Command/General/MediumMigrator.php
kariae: @naxoc it should be `sanitize_user($author['display_name'])` as the field should be sanitized, right?

src/Command/General/MediumMigrator.php
kariae: > which is not as good as our standard one, and will create dupes.

That's why I got some code from that repository and combined it with our `import_external_file ()` :D

src/Command/General/MediumMigrator.php
kariae: Nice catch!

src/Command/General/MediumMigrator.php
kariae: Isn't this covered by [this line](https://github.com/Automattic/newspack-custom-content-migrator/pull/396/files#diff-691ca5f3aad12c7079a7181ff55132bb539d2c96f598b5aa61aaea5f2c993976R1075)?

naxoc: The line in the link is the user login. It's the nicename we should set. I think it's unnecessary to sanitzie it – WP does that.

kariae: `user_nicename` is a sub-string of `user_login` if not set ([source code](https://github.com/WordPress/wordpress-develop/blob/87dfd5514b52aef456b7232b1959873e69e651da/src/wp-includes/user.php#L2153)). I don't think display names from Medium.com can be set as email addresses, but I can add a check https://github.com/Automattic/newspack-custom-content-migrator/pull/396/commits/4b8c62b0cd8b437761741c852deeccb676de0e30.

naxoc (approved):
Looks good! 
 OK

PR #399, opened 2024-01-03
Remove unused html purifier library

naxoc:
This library was promising, but ended up causing more trouble than helping. This removes it. `simplehtmldom\HtmlDocument` ended up being the way to go.

---

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -29,8 +29,7 @@
     "automattic/newspack-content-converter": "dev-master",
     "halaxa/json-machine": "^1.1",
     "mustache/mustache": "^2.14",
-    "simplehtmldom/simplehtmldom": "2.0-RC2",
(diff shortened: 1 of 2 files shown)


kariae:
It's used in [the HCN migrator](https://github.com/Automattic/newspack-custom-content-migrator/blob/master/src/Command/PublisherSpecific/HighCountryNewsMigrator2.php) I'm just not sure if this is the one we're using for this publisher 😅

naxoc:
Woops, sorry @kariae – it's removed in the [branch I'm actually working on for HCN](https://github.com/Automattic/newspack-custom-content-migrator/pull/382). I should have commented on that :) I can merge that branch first if you prefer?

kariae (approved):
All good 👍

PR #400, opened 2024-01-03
Initial commit of ConsoleColor to facilitate adding color to output!

eddiesshop:
They say, action speak louder than words. So let me show you this class in action:
image

Cool, right?!

Here's another useful one:
image

Features of this class:
1. This class uses magic methods to handle function calls dynamically. This is what allows you to use it from a static context (e.g. `ConsoleColor::bright_white('Testing, testing')->output();`)
2. Every call to a function which applies color or style will return an instance of the class, which is what allows you to build a string with different effects applied to different sections (i.e. `ConsoleColor::underlined_bright_cyan_with_blue_background('Live long')->white('and')->underlined_bright_white('prosper!'`)
3. Since method calls are handled dynamically, you could technically create your own custom color combination (text color, style, and background), but I've used ChatGPT to come up with a lot of good ones already (i.e. the PHPDoc block with all the `@method` tags). This will help IDE's with autocomplete when you use this class.
4. All text is sanitized on the way in (via the `apply()` function)
5. You can apply different text color, styles, and background to different sections of text. Only now, it'll look much cleaner and discernible!
6. Can be used in non-static context as well. Simply new-up the class. Every time you call `get()` or `output()` the colored string you're building will be returned and reset.


src/Utils/ConsoleColor.php
```diff
@@ -0,0 +1,739 @@
...
+	 * This function applies a color, style, or background to a string.
+	 *
+	 * @param string $type Whether to apply a color, style, or background.
+	 * @param string $property The color, style, or background to apply.
+	 * @param string $text The string to apply the color, style, or background to.
+	 *
+	 * @return string
+	 */
+	private function apply( string $type, string $property, string $text ): string {
+		$text_sanitized = wp_kses( $text, wp_kses_allowed_html( 'post' ) );
```
claudiulodro, line 731: Since the terminal doesn't use HTML, [sanitize_textarea_field](https://developer.wordpress.org/reference/functions/sanitize_textarea_field/) or [sanitize_text_field](https://developer.wordpress.org/reference/functions/sanitize_text_field/) are probably the "best" choice for sanitizing, but this shouldn't cause any issues.

claudiulodro (approved):
Handy feature! I like the way it does the chaining. Code looks solid, I just left a little feedback around the sanitizing function.

src/Utils/ConsoleColor.php
claudiulodro: Although now I do wonder if it **does** make sense to let it do HTML, so we can do e.g. color-coded diffs of post content or other things that are in HTML.

eddiesshop: Yea that was my thinking, that we'd want to be as permissive as possible, while still doing some sanitizing, because we work with so many different types of data sets, and some may contain HTML code in it

PR #401, opened 2024-01-05
feat(WindyCityMigrator): add migrator and fix YouTube GB block

kariae:
This is a PR for a publisher migrator. I can provide a data sample on Slack :) 

This PR also fixes the `get_youtube` from the Gutenberg Block Generator where before the fix the method deleted the $_GET parameters, which was deleting the `v=video_id` parameter.

### DIFF
--- a/src/Command/PublisherSpecific/WindyCityMigrator.php
+++ b/src/Command/PublisherSpecific/WindyCityMigrator.php
@@ -126,7 +126,7 @@ class WindyCityMigrator implements InterfaceCommand {
 		$this->logger->log( self::LOG_FILE, sprintf( 'Migrating %d entries.', $total_entries ) );
 
 		foreach ( $entries as $index => $entry ) {
-			$this->logger->log( self::LOG_FILE, sprintf( 'Migrating entry %d/%d.', $index + 1, $total_entries ), Logger::LINE );
+			$this->logger->log( self::LOG_FILE, sprintf( 'Migrating entry %d/%d.', $index + 1, $batch_args['end'] ), Logger::LINE );
 
 			// Check if post exists.
 			if ( in_array( $entry['GUID'], $existing_original_ids ) ) {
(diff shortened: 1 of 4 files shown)


src/Command/PublisherSpecific/WindyCityMigrator.php
```diff
@@ -0,0 +1,398 @@
...
+	 *
+	 * @param string $display_name Author name.
+	 * @param string $default_author_display_name Default author display name.
+	 * @param string $default_author_email Default author email.
+	 *
+	 * @return int|bool Author ID or false on failure.
+	 */
+	private function get_create_author( $display_name, $default_author_display_name, $default_author_email ) {
+		if ( empty( $display_name ) ) {
+			$username = sanitize_user( $default_author_display_name, true );
```
eddiesshop, line 245: Just learned about a new function! The only comment I have here is that I believe we should strive to obfuscate what a user's login and user_nicename would be. If we use Display Name to generate a login, that is exposed to the front end if someone goes to `{site}/author/{user_nicename}`. Perhaps we can use Display Name only for user_nicename generation, and use something like the first half of an email (before the @ symbol) as the login?

kariae: I added https://github.com/Automattic/newspack-custom-content-migrator/pull/401/commits/28ae5218c3711bc5e88163b6054a4331ce9a1831 to not expose the email address in both fields, does it work?

eddiesshop: Zak and I spoke about this change and I understand a bit more about the context. Basically this is the best we could do because of the data we're provided by the publisher.

This publisher needs to get and/or create a default author when their import doesn't specify one. 

Then for the case where we do have a `$display_name` specified, we don't have an email we could use to uniquely identify this author with, so we're relying on `user_login`/`user_nicename` for that.

Marking as resolved 👍🏼

eddiesshop (approved):

PR #404, opened 2024-01-10
fix(Attachments): fix extension of images without one

kariae:
Using `wp_get_mime_types()` to get the mime type  of an image returns `jpg|jpeg|jpe` as a key to the JPEG mime type instead of one extension, which breaks `media_handle_sideload` since we'll be uploading a file with a `name` ending with `file_name.jpg|jpeg|jpe`. This is a little fix for it.

## How to test
- Use `import_external_file` to upload a JPG image with a filename that doesn't contain the extension.
- Notice the upload is not working.
- Apply the patch and re-run the import.
- Notice that the upload is working.

---

### DIFF
--- a/src/Logic/Attachments.php
+++ b/src/Logic/Attachments.php
@@ -98,6 +98,12 @@ class Attachments {
 			// Without the extension, the upload will fail because WP will not allow that "file type".
 			$mimetype           = mime_content_type( $tmpfname );
 			$probably_extension = array_search( $mimetype, wp_get_mime_types() );
+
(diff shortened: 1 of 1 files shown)


naxoc (approved):
Awesome. I've been seeing errors from this lately. We won't always be right about the first one in the exploded array, but it's a lot better than nothing.

kariae:
I guess if we set whatever extension from the explode it will do the trick since it's the same mime type.

naxoc:
> I guess if we set whatever extension from the explode it will do the trick since it's the same mime type.

Good point. I was thinking about the call to see if we already have the file. But either way this fixes it!

PR #406, opened 2024-01-12
Adds PHP constant to configure error reporting level

iuravic:
Introduces a PHP constant which can change the error reporting level.

## How to test
- add constant with value to `wp-config.php`:
`define( 'NEWSPACK_CUSTOM_CONTENT_MIGRATOR_ERROR_REPORTING_LEVEL', 'live' );`
- confirm that error reporting level is not changed

---


newspack-custom-content-migrator.php
```diff
@@ -20,8 +20,8 @@
 }
 
 require_once ABSPATH . 'wp-settings.php';
-
-PluginSetup::configure_error_reporting();
+$error_reporting_level = false !== defined( 'NEWSPACK_CUSTOM_CONTENT_MIGRATOR_ERROR_REPORTING_LEVEL' ) ? NEWSPACK_CUSTOM_CONTENT_MIGRATOR_ERROR_REPORTING_LEVEL : 'dev';
```
leogermani, line 23: nit: the `false !==` only makes it more difficult to read this line

leogermani: WDYT of simply reading WP_DEBUG instead of having a new constant?

iuravic: @leogermani , I just thought of the reason why tying the error reporting verbosity to WP_DEBUG constant will not be good enough.

The main idea behind this default high verbose error output level was that developers (in this case we're talking mostly about LEs, since this is our core migration work plugin) have had different custom development environment setups -- some had all warnings and errors outputted while others had them suppressed on their env configs. This ensured that all devs saw all the warnings.

Since this is a dev plugin, we are interested in always having all warnings visible. And this plugin is not intended to be run on live -- that should be an exception. Since not all dev env configs will have WP_DEBUG turned ON by default, we won't get the desired verbose behavior by default.

I would like to leave this new constant implementation -- if it's set to 'live', then extra warnings are suppresed, otherwise verbose warnings get output by default.

leogermani: Sure! It makes sense!

PR #407, opened 2024-01-15
fix(EmbarcaderoMigrator): add text styling and fix broken URLs in content

kariae:
This PR fixes broken URLs migrated in content and migrates text styling.

## How to test
- Search in the database posts you migrated with broken migrated URLs `post_content like '%>http<%'`.
- Check the post frontend and notice that the article contains some URLs with `http` as text.
- Get the post-original ID from the database using the meta `_newspack_import_id`.
- Add a condition around line 547 to migrate only that post (only for test purposes), something like:
```php
if ('123456' !== $post['story_id'] ) {
    continue;
}
```

- Run the migrator with the flags `--refresh-content --skip-post-media` to refresh the content without downloading the images.
- Recheck the post and notice that the URLs are fixed.
- Do the same for the text styling tags:
  - Italic: `post_content like '%==I%'`
  - Bold: `post_content like '%==B%'`
  - Bold & Italic: `post_content like '%==BI%'`
---

### DIFF
--- a/src/Command/General/EmbarcaderoMigrator.php
+++ b/src/Command/General/EmbarcaderoMigrator.php
@@ -473,6 +473,23 @@ class EmbarcaderoMigrator implements InterfaceCommand {
 		);
 
 		WP_CLI::add_command(
+			'newspack-content-migrator embarcadero-helper-validate-csv-file',
+			array( $this, 'cmd_embarcadero_helper_validate_csv_file' ),
+			[
+				'shortdesc' => 'A helper command which validates a CSV file and outputs rows with issues.',
+				'synopsis'  => [
+					[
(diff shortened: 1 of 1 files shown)


iuravic (approved):
I've tested several posts for Pleasanton, from each of the `like` cases, e.g.
- `post_content like '%>http<%'`: postID 2944, storyID 6820
- `post_content like '%==BI%'`: postID 14186, storyID 12856
- `post_content like '%==I%'`: postID 11660, storyID 11453

Things seem to be transformed correctly 👍 

I've run into one example that seems to have a wrong link in the CSV. Please just verify that this is a source data error, not our script's error:
image

PR #408, opened 2024-01-17
Add a "block-transformer" that "hides" blocks for the NCC

naxoc:
This is a bit of an experimental attempt at getting mixed classic and GB blocks to survive an NCC run.

The idea is that we base64-encode all GB blocks before running NCC and then decode them after NCC is done.

It's for specific use for a publisher now and certainly needs some docs and polishing, but it should be kinda resuable. I'm also not sure I love the name of the class, so we can change that if we find a better name.

## How to test
- To encode, run `wp newspack-content-migrator transform-blocks-encode --post-id=XXX`
- Run NCC on the post
- Decode with `wp newspack-content-migrator transform-blocks-decode --post-id=XXX`

The post should have survived without being chopped to pieces by NCC or by bad decoding.

---


iuravic (changes requested):
This logic looks really nice and clean to me 👍  And it works great :high-five! :) 

I think it makes a lot of sense to incorporate this mechanism into NCC release. In order to do that more easily (because there's a bunch of pure NCC work to be done in order to release it), it might help a lot to just refactor this functionality a little bit.

I wonder what would you think of the following proposals:
- creating a new class Logic\GutenbergBlockTransformer
- inside the new Logic class, creating 4 methods:
  - `encode_block` -- the existing one from this PR and Command class
  - `encode_post_content( $html ) : $html_encoded` -- one that takes post_content HTML, parses blocks, and calls encode_block
  - `decode_block` -- existing one
  - `decode_post_content( $html_encoded) : $html`

I understand if you perhaps would be hesitant to do such a refactor, because this code already works great! But in case it wouldn't be too much of a bother, we can simply contribute this part to the NCC v.1 release.

Lastly, if we could run PHPCS before merge that would be awesome :)

naxoc:
Thanks, @iuravic !

I refactored a little bit so there are now two classes. Please take a look again :)

iuravic (approved):

PR #409, opened 2024-01-17
feat(EmbarcaderoMigrator): allow only a set of categories

kariae:
## How to test
- Run the new command `newspack-content-migrator embarcadero-rearrange-categories`
- Notice the categories set (all the categories should belong to the list from the constant `ALLOWED_CATEGORIES`, or the `General` category).

---


src/Command/General/EmbarcaderoMigrator.php
```diff
@@ -74,6 +74,111 @@ class EmbarcaderoMigrator implements InterfaceCommand {
 	const EMBARCADERO_ORIGINAL_MEDIA_ID_META_KEY    = '_newspack_media_import_id';
 	const DEFAULT_AUTHOR_NAME                       = 'Staff';
 
+	const ALLOWED_CATEGORIES = [
```
eddiesshop, line 77: For readability this is great, but if it's not too much of a pain, let's convert these to lowercase so that we won't need to convert them during runtime. I can create a commit for this if you'd like.

src/Command/General/EmbarcaderoMigrator.php
```diff
@@ -912,6 +1064,69 @@ public function cmd_fix_post_times( array $args, array $assoc_args ): void {
...
+					sprintf( 'Entry not found %s.', $post['story_id'] ),
+					$this->logger::WARNING
+				);
+
+				continue;
+			}
+
+			// Set categories from sections data.
+			$imported_category  = '';
+			$post_section_index = array_search( $post['section_id'], array_column( $sections, 'section_id' ) );
```
eddiesshop, line 1105: Can we pull out `array_column( $sections, 'section_id' )` from the loop instead?

eddiesshop (changes requested):
Aside from 2 small comments, this looks good to me.

src/Command/General/EmbarcaderoMigrator.php
kariae: I don't think the `strtolower` will break efficiency in this case, but got them lowercased here https://github.com/Automattic/newspack-custom-content-migrator/pull/409/commits/0d89b0f86442c4142870bda084d31b3376478c75 :)

src/Command/General/EmbarcaderoMigrator.php
eddiesshop:
```suggestion
			$post_section_index = array_search( $post['section_id'], $section_ids );
```

src/Command/General/EmbarcaderoMigrator.php
```diff
@@ -912,6 +1062,67 @@ public function cmd_fix_post_times( array $args, array $assoc_args ): void {
...
+	 */
+	public function cmd_embarcadero_rearrange_categories( $args, $assoc_args ) {
+		$story_csv_file_path          = $assoc_args['story-csv-file-path'];
+		$story_sections_csv_file_path = $assoc_args['story-sections-file-path'];
+		$index_from                   = isset( $assoc_args['index-from'] ) ? intval( $assoc_args['index-from'] ) : 0;
+		$index_to                     = isset( $assoc_args['index-to'] ) ? intval( $assoc_args['index-to'] ) : -1;
+
+		$posts    = $this->get_data_from_csv_or_tsv( $story_csv_file_path );
+		$sections = $this->get_data_from_csv_or_tsv( $story_sections_csv_file_path );
+
```
eddiesshop, line 1079:
```suggestion
			$section_ids = array_column( $sections, 'section_id' );
```

eddiesshop (commented):
Something like this instead

src/Command/General/EmbarcaderoMigrator.php
kariae: Done :)

eddiesshop (approved):

PR #410, opened 2024-01-17
fix(EmbarcaderoMigrator): fix More Posts block

kariae:
## How to test
- Run this new command `newspack-content-migrator embarcadero-fix-more-posts-block`
- Check some posts from the logs that contain an aligned to the right Group Block that contains "More Posts" posts.

---

### DIFF
--- a/src/Command/General/EmbarcaderoMigrator.php
+++ b/src/Command/General/EmbarcaderoMigrator.php
@@ -311,6 +311,30 @@ class EmbarcaderoMigrator implements InterfaceCommand {
 		);
 
 		WP_CLI::add_command(
+			'newspack-content-migrator embarcadero-fix-more-posts-block',
+			array( $this, 'cmd_embarcadero_fix_more_posts_block' ),
+			[
+				'shortdesc' => 'Fix Embarcadero\s post "more posts" block.',
+				'synopsis'  => [
+					[
+						'type'        => 'assoc',
+						'name'        => 'story-csv-file-path',
+						'description' => 'Path to the CSV file containing the stories to import.',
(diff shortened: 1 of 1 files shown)


src/Command/General/EmbarcaderoMigrator.php
```diff
@@ -990,6 +1016,118 @@ function( $media_item ) use ( $post ) {
...
+
+				// Find the 'More stories' heading block index.
+				$more_stories_heading_block_index = array_search(
+					$this->gutenberg_block_generator->get_heading( 'More stories' )['innerHTML'],
+					array_column( $content_blocks, 'innerHTML' ),
+					true
+				);
+
+				// Check if the post contains the "More stories" heading block.
+				if ( false === $more_stories_heading_block_index ) {
```
eddiesshop, line 1090: Should we perhaps flag this a little more prominently, since it's established that there is at least something in `post_content` with the text `>More stories<`? Maybe put the message in ALL CAPS?

eddiesshop (approved):

src/Command/General/EmbarcaderoMigrator.php
kariae: That's a choice we can make for all warning messages as technically we need to check all warning messages from the log file once a migration command is run. I filter the log file lines that start with `Warning:` and check them. Shall we uppercase all the warnings?

eddiesshop: Nah, if you're cool with it being that way, I don't mind. You're reasoning makes sense, and it's more a personal preference thing. All good 👍

PR #412, opened 2024-01-17
Fix/csv file class

eddiesshop:
## How to test
Get your most problematic CSV file, and see if this class will read it and process the rows.
```
$ wp shell
> $csv = ( new FileImportFactory )->get_file( '/path/to/file' )->getIterator();
> foreach ($csv as $row ) {
>     var_dump( $row );
> }
```
---

_**I think**_ [this is the best way](https://github.com/Automattic/newspack-custom-content-migrator/blob/fix/csv-file-class/src/Utils/CommonDataFileIterator/CSVFile.php#L131) to handle charsets, but would appreciate if someone more knowledgeable than me could check or confirm.


naxoc (approved):
I tested this on a semi-wonky csv and it worked great!

This looks good to go. 

As for handling the charset stuff, I think I'd prefer it to have it be an arg, so that a human has looked at encoding and made that decisison. I think falling back to UTF-8 could cause trouble. None of that is blocking this PR though – it's more of a personal preference. :)

eddiesshop:
Thanks for the review @naxoc!

That makes sense, I'll create another PR adding that flexibility.

PR #413, opened 2024-01-19
feat(EmbarcaderoMigrator): add batching for cmd_fix_post_times

kariae:
## How to test
- Run the `newspack-content-migrator embarcadero-helper-fix-dates` with the parameters `index-from` and `index-to`
- Notice in the logs it's only running the batch we set

---

### DIFF
--- a/src/Command/General/EmbarcaderoMigrator.php
+++ b/src/Command/General/EmbarcaderoMigrator.php
@@ -655,6 +655,20 @@ class EmbarcaderoMigrator implements InterfaceCommand {
 						'optional'    => false,
 						'repeating'   => false,
 					],
+					[
+						'type'        => 'assoc',
+						'name'        => 'index-from',
+						'description' => 'Start importing from this index in the CSV file.',
+						'optional'    => true,
+						'repeating'   => false,
+					],
+					[
+						'type'        => 'assoc',
+						'name'        => 'index-to',
+						'description' => 'Import till this index in the CSV file.',
+						'optional'    => true,
+						'repeating'   => false,
+					],
 				],
 			]
 		);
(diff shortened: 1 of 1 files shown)


src/Command/General/EmbarcaderoMigrator.php
```diff
@@ -1085,12 +1099,21 @@ private function get_post_date_from_timestamp( int $timestamp ): string {
 	 */
 	public function cmd_fix_post_times( array $args, array $assoc_args ): void {
 		$story_csv_file_path = $assoc_args['story-csv-file-path'];
+		$index_from          = isset( $assoc_args['index-from'] ) ? intval( $assoc_args['index-from'] ) : 0;
+		$index_to            = isset( $assoc_args['index-to'] ) ? intval( $assoc_args['index-to'] ) : -1;
+
 
 		$posts       = $this->get_data_from_csv_or_tsv( $story_csv_file_path );
 		$log_file    = 'fix-post-times.log';
 		$total_posts = count( $posts );
```
naxoc, line 1108: Could we move this line under the next conditional? The total is wrong in logging because it counts before slicing :)

src/Command/General/EmbarcaderoMigrator.php
```diff
@@ -1085,12 +1099,21 @@ private function get_post_date_from_timestamp( int $timestamp ): string {
...
+
+		// Get selected posts.
+		if ( -1 !== $index_to ) {
+			$posts = array_slice( $posts, $index_from, $index_to - $index_from + 1 );
+		}
+
 		foreach ( $posts as $post_index => $post ) {
-			$this->logger->log( $log_file, sprintf( 'Importing post %d/%d: %d', $post_index + 1, $total_posts, $post['story_id'] ), Logger::LINE );
+			$this->logger->log( $log_file, sprintf( 'Importing post %d/%d: %d', $index_from + $post_index + 1, $total_posts, $post['story_id'] ), Logger::LINE );
```
naxoc, line 1116: Small nit – feel free to ignore: I know I wrote this, but can we change "importing" to "fixing timezone". I managed to make myself confused when I ran it 😆

naxoc (approved):
I had a small request and an even smaller nit, but I tested and it batches neatly.

src/Command/General/EmbarcaderoMigrator.php
kariae: Yup, I thought showing the total posts was a better hint (I know it's not the case for the rest of the commands), I'll do so.

PR #418, opened 2024-01-24
feat(EmbarcaderoMigrator): add command to fix text styling

kariae:
## How to test
- Run the command `embarcadero-fix-content-styling` locally on all your Embarcadero site content using the parameter `posts-per-batch=-1`
- Check the updated posts from the logs.
- Check that no more text needs styling by running the following SQL commands (the count should be 0):
```sql
SELECT COUNT(ID) FROM `wp_posts` WHERE post_type = 'post' AND post_status = 'publish' AND post_content LIKE '%==I%';
SELECT COUNT(ID) FROM `wp_posts` WHERE post_type = 'post' AND post_status = 'publish' AND post_content LIKE '%==B%';
SELECT COUNT(ID) FROM `wp_posts` WHERE post_type = 'post' AND post_status = 'publish' AND post_content LIKE '%==BI%';
SELECT COUNT(ID) FROM `wp_posts` WHERE post_type = 'post' AND post_status = 'publish' AND post_content LIKE '%==SH%';
```

---


naxoc (approved):
Tested and looks good. It brought down the number of posts with `post_content LIKE '%==I%';` for me :)

PR #421, opened 2024-01-29
feat(MediumMigrator): set tags as categories

kariae:
This PR sets the Medium's posts migrated tags as categories instead of post tags. Since Medium doesn't have the notion of categories and we use categories for our articles structure instead of tags.

## How to test
- Get a Medium export, I can provide that.
- On an empty WP site, run the migration command `wp newspack-content-migrator migrate-medium-archive --zip-archive=/path/to-the/archive/file.zip`
- Notice the list of tags in the dashboard is empty.
- Notice the list of categories containing some new ones, that are linked to the imported posts.

---

### DIFF
--- a/src/Logic/Medium.php
+++ b/src/Logic/Medium.php
@@ -98,14 +98,14 @@ class Medium {
 	const MISSING_POST_URL = "The original URL couldn't be found in exported file.";
 
 	/**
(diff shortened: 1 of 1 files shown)


naxoc (approved):
Looks good! 
 OK

PR #422, opened 2024-01-31
feat(WindyCityMigrator): migrate post's PDFs

kariae:
This PR fix migrating galleries and add the feature of migrating PDFs for Windy City

---


naxoc (approved):
I read through this code and it looks good to me
 OK

PR #423, opened 2024-01-31
Feat/migration helper parameters with space

kariae:
This PR makes it possible to have commands with space in their parameter values and fix the counting of the `end` parameter.

## How to test
- Run the following command `wp newspack-content-migrator migration-helper-output-batched 'wp newspack-content-migrator my-custom-migrator --param1=123 --param2="filename with space.zip"' --total-items=20 --batch-size=2 --begin-at=5`
- Notice the generated commands are breaking for the parameter with space `param2`
- Notice the three last commands don't have an `end` parameter.
- Pull this branch.
- Run the same command.
- Notice that the commands are generated with the spaced parameter on one line.
- Notice the `end` parameter is set to all the commands.

---

### DIFF
--- a/src/Command/General/MigrationHelper.php
+++ b/src/Command/General/MigrationHelper.php
@@ -8,11 +8,10 @@ use NewspackCustomContentMigrator\Utils\BatchLogic;
 use WP_CLI;
 use WP_CLI\ExitException;
 
+/**
+ * Class MigrationHelper.
+ */
 class MigrationHelper implements InterfaceCommand {
-	private function __construct() {
-		// Do nothing.
-	}
-
 	/**
 	 * Get Instance.
 	 *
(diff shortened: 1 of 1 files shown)


src/Command/General/MigrationHelper.php
```diff
@@ -107,38 +115,46 @@ public function cmd_output_batched( array $positional_args, array $assoc_args ):
...
-		$num_batches = ceil( $total_items / $batch_size );
+		$total_batched_items = $begin_at > 0 ? $total_items - $begin_at : $total_items;
+		$num_batches         = ceil( $total_batched_items / $batch_size );
 
 		for ( $batch = 0; $batch < $num_batches; $batch ++ ) {
 			$batch_start = ( $batch * $batch_size ) + $begin_at;
 			$batch_end   = $batch_start + $batch_size;
 			if ( $batch_end > $total_items ) {
-				$batch_end = 0;
+				$batch_end = $total_items + 1;
```
naxoc, line 130: This counts wrong? `--end` is inclusive in the iterator helpers, so I'm not sure we should add one?

kariae: @naxoc nice catch, I pushed a fix for it!

naxoc (approved):
Looks good! 
 OK

PR #426, opened 2024-02-01
Rename attachment images to be supported by Subsizes

iuravic:
## This PR

Subsizes does not allow attachment filenames to end in -\d+x\d+, e.g. `image-100x200.jpg`. If you upload a file like that to Atomic, it will be accepted, but a `-1` will be appended to its filename and attachment post object's post_name, e.g. you will get `image-100x200-1.jpg` uploaded to your media library. If we're cloning historic WP sites to Atomic, such images won't work once Subsizes starts appending its size suffixes.

This PR fixes those attachments by:
- looping through all attachment objects and find which end in -\d+x\d+ suffix
- renames those files and attachment objects
- updates usage of those attachments in post_content

## Caveat!

This needs to be run before conversion to blocks. Because conversion to blocks might append an additional -\d+x\d+ on top of the existing old -\d+x\d+. E.g. invalid image in HTML `Dorothy-336x225.jpg` might be updated to `Dorothy-336x225-336x225.jpg` by Gutenberg. So it's important to run this before converting to blocks.

## How to test
- create a site on your Local, and upload some local images with suffixes, like `image-100x200.jpg`
- use these images in posts
- run the command
- verify that `image-100x200.jpg` has been renamed to `image-100x200-1.jpg` both in Media Library's URL to image, and in posts where this image was used

---

### DIFF
--- a/src/Command/General/AttachmentsImagesSubsizesMigrator.php
+++ b/src/Command/General/AttachmentsImagesSubsizesMigrator.php
@@ -1,0 +1,239 @@
+<?php
+
+namespace NewspackCustomContentMigrator\Command\General;
+
+use \NewspackCustomContentMigrator\Command\InterfaceCommand;
+use \NewspackCustomContentMigrator\Logic\Posts;
+use \WP_CLI;
+
+class AttachmentsImagesSubsizesMigrator implements InterfaceCommand {
+
+	/**
+	 * Instance.
+	 *
+	 * @var null|InterfaceCommand Instance.
+	 */
+	private static $instance = null;
+
+	/**
+	 * Posts instance.
+	 *
+	 * @var Posts Posts instance.
+	 */
+	private $posts;
+
+	/**
+	 * Constructor.
+	 */
+	private function __construct() {
+		$this->posts = new Posts();
+	}
+
+	/**
+	 * Singleton get_instance().
+	 *
+	 * @return InterfaceCommand|null
+	 */
+	public static function get_instance() {
+		$class = get_called_class();
+		if ( null === self::$instance ) {
+			self::$instance = new $class();
+		}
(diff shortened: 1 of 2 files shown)


src/Command/General/AttachmentsImagesSubsizesMigrator.php
```diff
@@ -0,0 +1,247 @@
...
+	private function __construct() {
+		$this->posts = new Posts();
+	}
+
+	/**
+	 * Singleton get_instance().
+	 *
+	 * @return InterfaceCommand|null
+	 */
+	public static function get_instance() {
```
naxoc, line 45: This is a nit, but there is no reason to keep an instance var for the simpleton, so this would be simpler:
```suggestion
	public static function get_instance(): self {
		static $instance = null;
		if ( null === $instance ) {
			$instance = new self();
		}

		return $instance;
	}
```

naxoc (approved):
This looks great – if the `wp_unique_filename` won't work for this feel free to merge it.

The nit is not important, but I keep seeing the instance var in our code and it confuses me 😆 

Generally I think we should use PHP 8 return declarations on functions now that we have left PHP 7 behind, so I'd love to also see them. (feel free to also ignore that one).

iuravic:
> Generally I think we should use PHP 8 return declarations on functions now that we have left PHP 7 behind, so I'd love to also see them. (feel free to also ignore that one).

Agreed, I think we're ready to add this to our PHPCS standard. Feel free to if you'd like.

src/Command/General/AttachmentsImagesSubsizesMigrator.php
iuravic: Updated.

PR #433, opened 2024-02-12
fix(Medium): fix importing embeds (and add Frisc)

kariae:
Custom publisher migrator with a command to clean content.

---


naxoc:
I got a bit lost in the branch renaming to trunk and ended up pushing to this PR. It's not _all_ wrong, but a little off. I'm also adding a Frisc migrator in this one on top of Zak's stuff.

PR #453, opened 2024-03-12
Fix/town news migator ga creation error

eddiesshop:
This change fixes some runtime issues that were experienced while importing content for Knox Pages.

---


iuravic (approved):

eddiesshop:
Hey @naxoc ! Since that one is a little stale, I went ahead and merged this one, but I think we should add the changes from that PR also. If you don't mind making the updates, I promise to review the changes!

PR #457, opened 2024-03-18
Add a command to fix posts imported with wrong timezone

naxoc:
## How to test
- Create a post (or edit one in the DB) to have a date in some timezone
- Run converter
- Check that the timezone was fixed.

Run like this for example: 
`wp newspack-content-migrator change-posts-timezone --from-timezone=UTC --target-timezone='America/Chicago' --num-items=500`

---


naxoc:
Thanks for the feedback @iuravic . Good points! I've pushed those things, so it's ready for re-review.

naxoc:
Oh, just saw you didn't need a second look. Merging! Sorry 😆

iuravic (approved):

PR #460, opened 2024-03-20
LA Focus: Migration

jorostoyanov:
## How to test

---

### DIFF
--- a/src/Command/General/InlineFeaturedImageMigrator.php
+++ b/src/Command/General/InlineFeaturedImageMigrator.php
@@ -365,6 +365,27 @@ class InlineFeaturedImageMigrator implements InterfaceCommand {
 		foreach ( $posts_wo_featured_img as $k => $post ) {
 			WP_CLI::line( sprintf( '👉 (%d/%d) ID %d ...', $k + 1, count( $posts_wo_featured_img ), $post->ID ) );
 
+			// Check if the post content is generated using blocks and search for the first core/image block
+			// If such block is found, use it to set the Featured Image
+			if ( has_block( 'image', $post ) ) {
+				WP_CLI::line( '✓ found Gutenberg image block.' );
+
+				$post_blocks = parse_blocks( $post->post_content );
+				$post_core_image_blocks = array_filter( $post_blocks, function ( $block ) {
+					return $block['blockName'] === 'core/image' && ! empty($block['attrs']['id']);
+				} );
+
+				if ( ! empty( $post_core_image_blocks ) ) {
+					foreach ( $post_core_image_blocks as $image_block ) {
+						if ( ! empty( $image_block['attrs']['id'] ) ) {
+							set_post_thumbnail( $post, $image_block['attrs']['id'] );
+							WP_CLI::line( sprintf( '✓ Updated Featured Image from core/image Gutenberg block. ID %d', $image_block['attrs']['id'] ) );
+							continue 2;
+						}
+					}
+				}
+			}
+
 			// Find the first <img>.
 			$crawler->clear();
 			$crawler->add( $post->post_content );


src/Command/General/InlineFeaturedImageMigrator.php
```diff
@@ -508,4 +525,22 @@ public function cmd_remove_featured_images_from_beginning_of_postcontent( $args,
...
+	 * Returns an array of all "core/image" blocks from the given
+	 * array of post content blocks.
+	 */
+	private function get_image_blocks_from_post_content_blocks( array $post_content_blocks ): array {
+		$image_blocks = [];
+
+		foreach ( $post_content_blocks as $block ) {
+			if ( $block['blockName'] === 'core/image' && ! empty( $block['attrs']['id'] ) ) {
+				$image_blocks[] = $block;
+			} else if ( ! empty( $block['innerBlocks'] ) ) {
```
naxoc, line 539: Nice! We haven't thought much about nested blocks in our block handlers, so good call here!

src/Command/PublisherSpecific/LAFocusMigrator.php
```diff
@@ -0,0 +1,52 @@
...
+ * Custom migration scripts for LA Focus.
+ */
+class LAFocusMigrator implements InterfaceCommand {
+
+	/**
+	 * Singleton.
+	 *
+	 * @return LAFocusMigrator
+	 */
+	public static function get_instance(): self {
```
naxoc, line 24: Not a big deal, but we'd need a private constructor to enforce the singleton.

src/Command/PublisherSpecific/LAFocusMigrator.php
```diff
@@ -0,0 +1,52 @@
...
+			[
+				'shortdesc' => 'Searches for the first image in Post Content and uses is a Featured Image in full size.',
+			]
+		);          
+	}
+
+	/** 
+	 * Searches for the first image in Post Content and uses it as a Featured Image in full size.
+	 */
+	public function cmd_lafocus_set_featured_image_from_first_image_in_post_content( array $args, array $assoc_args ): void {
```
naxoc, line 49: Small thing, but since we are already in a namespaced class, we don't need the publisher name in functions.

src/Command/PublisherSpecific/LAFocusMigrator.php
```diff
@@ -0,0 +1,52 @@
...
+		WP_CLI::runcommand( 'newspack-content-migrator set-first-image-from-content-as-featured-image' );
```
naxoc, line 50: We need to add `[ 'launch' => false ]` to run it in the same process.

src/Command/General/InlineFeaturedImageMigrator.php
```diff
@@ -365,6 +365,23 @@ public function cmd_set_first_image_from_content_as_featured_image( $args, $asso
...
 			WP_CLI::line( sprintf( '👉 (%d/%d) ID %d ...', $k + 1, count( $posts_wo_featured_img ), $post->ID ) );
 
+			// Check if the post content is generated using blocks and search for the first core/image block
+			// If such block is found, use it to set the Featured Image
+			$image_blocks = $this->get_image_blocks_from_post_content_blocks( parse_blocks( $post->post_content ) );
+
+			if ( ! empty( $image_blocks ) ) {
+				WP_CLI::line( '✓ found Gutenberg image blocks.' );
+
+				foreach ( $image_blocks as $image_block ) {
```
naxoc, line 375: Small thing, but `continue 2` is a bit hard to read here. Can we do it in a way that doesn't use that? We generally use `continue` a bunch, but not skipping multiple scopes.

src/Logic/Attachments.php
```diff
@@ -65,6 +65,15 @@ public function import_attachment_for_post( int $post_id, string $path, string $
 	 * @return int|WP_Error Attachment ID.
 	 */
 	public function import_external_file( $path, $title = null, $caption = null, $description = null, $alt = null, $post_id = 0, $args = [], $desired_filename = '' ) {
+		// When using WP.com CDN, the URL can look like this:
+		// https://i0.wp.com/lafocus-newspack.newspackstaging.com/wp-content/uploads/2022/08/1_original_file_I0.jpg?resize=804%2C1024&ssl=1
+		//
+		// The ?resize=804%2C1024&ssl=1 part breaks the logic to download file and set proper extension and file data.
+		// So, we check if the URL matches a WP.com CDN pattern and strip the GET part.
+		if ( str_starts_with( $path, 'https://i0.wp.com' ) ) {
```
naxoc, line 73: This is not technically an external file. I'm not sure I agree that this should happen here. I think the code calling this should be responsible for checking if the url is external. 

Can we move it to InlineFeaturedImageMigrator maybe? We could add the code there (or a filter call for the url)  and check if the attachment exists given that. 

The CDN urls often start with `i0`, but the number can also be 1,,2, or maybe even 3. Don't remember :)

naxoc (changes requested):
thanks!

I left some small more code-style comments and one about the url juggling that I think should be moved.

src/Command/PublisherSpecific/LAFocusMigrator.php
jorostoyanov: Makes total sense 👍 I've added the private constructor 👌

src/Command/General/InlineFeaturedImageMigrator.php
jorostoyanov: Great thought 🙌 

I've now changed it to use a flag variable and continue statement outside the image blocks loop processing.

Overall, I think we might refactor the method structure in the future by using something like the [Chain of Responsibility Pattern](https://refactoring.guru/design-patterns/chain-of-responsibility) where each handler is a different method.

src/Logic/Attachments.php
jorostoyanov: ✅ I've moved it back to the `InlineFeaturedImageMigrator` class before even matching an image URL with an Attachment ID.

✅ The code now uses a regex to match any WP.com CDN pattern (i0, i1, i10..)

✅ In addition, I am now stripping the CDN prefix (in addition to the GET parameters) because I noticed that we can have a CDN URL for an external or internal attachment.

If the URL is internal it will find the Attachment ID and use it without uploading it again from the CDN (which was not the cause before this change)

If the URL is external it will use the `import_external_file` logic.

Hope this makes sense :)

naxoc (approved):
Looks great now. Thanks!

PR #475, opened 2024-04-04
Fix cmd_set_first_image_from_content_as_featured_image

iuravic:
### Fixes made to `cmd_set_first_image_from_content_as_featured_image`
- there is no need for the logic which separately works on blocks. The existing fetching of `` works well both on blocks and HTML
- removed this whole block, because the `$this->attachments->import_external_file` has a more advanced inner function where it will fetch existing attachment if it already exist -- [check this](https://github.com/Automattic/newspack-custom-content-migrator/blob/trunk/src/Logic/Attachments.php#L112). Also aways use `$this->attachments->import_external_file` when importing images in migration scripts since that's our team's standard
```
// Import attachment if it doesn't exist.
$att_id = 
    $attachment = get_post( $att_id );
    if ( $attachment ) {
        ...
```
- Updated comment to clarify what code was doing
```
fom: // The code below strips the CDN _pattern_ to leave the _original URL_ which can be both
to: // The code below strips the CDN **hostname and query parameters** and leaves the **local hostname without query params** which can be fetched both
```
- for the bit where query params are removed, I'd recommend using `parse_url` as it's are more readable (OK to leave this now 👍 )
- Added skip trying to set featured image if the current one is already correct

### Small updates to `cmd_set_first_image_from_content_as_featured_image`
- added optional argument to only process specific post IDs
- added optional argument to run on all posts, even those which already have a featured image
- initializing common logic objects in constructor, not in command
- I updated the logs to be more readable and **searchable**

## Publisher specific code
- ...

---

### DIFF
--- a/src/Command/General/InlineFeaturedImageMigrator.php
+++ b/src/Command/General/InlineFeaturedImageMigrator.php
@@ -2,13 +2,13 @@
 
 namespace NewspackCustomContentMigrator\Command\General;
 
-use \NewspackCustomContentMigrator\Command\InterfaceCommand;
-use \NewspackCustomContentMigrator\Logic\Attachments as AttachmentsLogic;
-use \NewspackCustomContentMigrator\Logic\Posts as PostLogic;
-use \NewspackCustomContentMigrator\Utils\Logger;
-use \WP_CLI;
-use \WP_Query;
-use Symfony\Component\DomCrawler\Crawler as Crawler;
+use NewspackCustomContentMigrator\Command\InterfaceCommand;
+use NewspackCustomContentMigrator\Logic\Attachments as AttachmentsLogic;
+use NewspackCustomContentMigrator\Logic\Posts as PostLogic;
+use NewspackCustomContentMigrator\Utils\Logger;
+use WP_CLI;
+use WP_Query;
+use Symfony\Component\DomCrawler\Crawler;
 
 /**
  * InlineFeaturedImageMigrator.
@@ -35,13 +35,29 @@ class InlineFeaturedImageMigrator implements InterfaceCommand {
 	 * @var Logger $logger Logger instance.
 	 */
 	private $logger;
+	
+	/**
+	 * Crawler.
+	 *
+	 * @var Crawler $crawler Crawler instance.
+	 */
+	private $crawler;
+
+	/**
+	 * AttachmentsLogic Instance.
+	 *
+	 * @var $attachments AttachmentsLogic.
+	 */
+	private $attachments;
 
 	/**
 	 * Constructor.
 	 */
 	private function __construct() {
-		$this->post_logic = new PostLogic();
-		$this->logger     = new Logger();
+		$this->post_logic  = new PostLogic();
+		$this->logger      = new Logger();
+		$this->crawler     = new Crawler();
+		$this->attachments = new AttachmentsLogic();
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


src/Command/General/InlineFeaturedImageMigrator.php
```diff
@@ -350,71 +380,52 @@ public function cmd_hide_featured_image_if_used_in_post_content( $pos_args, $ass
...
-				WP_CLI::line( sprintf( '✓ Updated Featured Image from core/image Gutenberg block. Attachment ID %d', $image_block['attrs']['id'] ) );
-				continue;
-			}
+		foreach ( $post_ids as $k => $post_id ) {
+			WP_CLI::line( sprintf( '(%d/%d) post ID %d ...', $k + 1, count( $post_ids ), $post_id ) );
+			
+			$current_thumbnail_id = get_post_thumbnail_id( $post_id );
+			
+			// Skip posts which already have a featured image.
+			if ( false == $re_set_existing_featured_images ) {
```
naxoc, line 403: Should this either be `===` or just `( ! $re_set_existing_featured_images )`?

src/Command/General/InlineFeaturedImageMigrator.php
```diff
@@ -423,37 +434,37 @@ public function cmd_set_first_image_from_content_as_featured_image( $args, $asso
...
+			if ( is_wp_error( $att_id ) ) {
+				WP_CLI::warning( sprintf( '❗ ERROR importing img `%s` to post ID %s : %s', $img_src, $post_id, $att_id->get_error_message() ) );
+				continue;
 			}
 
 			// Set attachment as featured image.
-			$result_featured_set = set_post_thumbnail( $post, $att_id );
-			if ( ! $result_featured_set ) {
-				WP_CLI::warning( sprintf( '❗ could not set att.ID %s as featured image', $att_id ) );
+			if ( $current_thumbnail_id == $att_id ) {
```
naxoc, line 453: Should be `===`

src/Command/General/InlineFeaturedImageMigrator.php
```diff
@@ -35,13 +35,29 @@ class InlineFeaturedImageMigrator implements InterfaceCommand {
 	 * @var Logger $logger Logger instance.
 	 */
 	private $logger;
+	
+	/**
+	 * Crawler.
+	 *
+	 * @var Crawler $crawler Crawler instance.
```
naxoc, line 42: So many crawler mentions! 😄

naxoc (approved):
Not sure if theis fixes something broken, but the code looks good.

I added some small comments.

jorostoyanov:
Hey @iuravic 

The reason why there was a separate blocks parsing is the following use-case. Imagine having the following block in the content:

```
<!-- wp:image {"align":"center","id":20174,"sizeSlug":"large","linkDestination":"none"} -->
<figure class="wp-block-image aligncenter size-large"><img src="https://example.com/wp-content/uploads/2024/02/sample-image-1024x538.jpg" alt="" class="wp-image-20174"/></figure>
<!-- /wp:image -->
```

There is an existing image `https://example.com/wp-content/uploads/2024/02/sample-image-1024x538.jpg`, but we don't want to use the resized version for a Featured Image.

The block contains the original Image ID (id: 20174) and we need to use it instead.

Otherwise, the command will upload the resized image again and use the new Attachment ID as Featured Image.

Hope that makes sense.

Let me know, if you need more examples :)

src/Command/General/InlineFeaturedImageMigrator.php
naxoc: I'm not sure I love having the crawler as an instance var come to think of it. Much clearer to read `new Crawler()` when one needs one and much less chance of foregetting to clear it.

PR #479, opened 2024-04-08
Add / Hide All Featured Images

ronchambers:
This general use CLI will add postmeta `newspack_featured_image_position = 'hidden'` to all posts that have a featured image.

Usage: `wp newspack-content-migrator hide-all-featured-images`

Output log: `hide_all_featured_images.log`

Note:
- There is a similar existing command (`newspack-content-migrator hide-featured-images`) in file (`src/Command/General/PostsMigrator.php`).  After reviewing that command it would have been more work to refactor it for this new use case.  The existing command requires a list of post ids and/or a category id.  Our use case doesn't call for a list of post ids nor category id.  Additionally the existing command needs to be called multiple times in order to process all rows (or use a very high `posts-per-batch` count that could take up too much processing power).  The new command in this PR makes use of `throttled_posts_loop` to by pass this concern.

---

- [x] **(only for this new code)** confirmed that PHPCS has been run

### DIFF
--- a/src/Command/General/InlineFeaturedImageMigrator.php
+++ b/src/Command/General/InlineFeaturedImageMigrator.php
@@ -129,6 +129,14 @@ class InlineFeaturedImageMigrator implements InterfaceCommand {
 				],
 			]
 		);
+
+		WP_CLI::add_command(
+			'newspack-content-migrator hide-all-featured-images',
+			[ $this, 'cmd_hide_all_featured_images' ],
+			[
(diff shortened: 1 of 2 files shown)


iuravic (changes requested):
This looks great 👍 
Might I trouble you to just do one change and move it to `src/Command/General/InlineFeaturedImageMigrator.php`? I totally understand that there are a few places where this functionality presently exists, but we've mostly been using this class for that lately.
Thank you 🙇

iuravic (approved):

PR #487, opened 2024-04-17
Add NCC trickery

naxoc:
This adds support for args in the block encode/decode commands:

```
  [--post-id=<post-id>]
    The ID of the post to process.

  [--num-items=<num-items>]
    The number of posts to process.

  [--min-post-id=<min-post-id>]
    The minimum post ID to process.

  [--max-post-id=<max-post-id>]
    The maximum post ID to process.

  [--post-types=<post-types>]
    Comma-separated list of post types to process. Default is "post".
```

There is a new command that "nudges" posts to be picked up by the NCC - it just inserts a line break at the top of posts that start with `<!--` that would otherwise not get picked up by the NCC.

## How to test

* Make sure the NCC is installed and activated
* Run `wp newspack-content-migrator transform-blocks-encode --num-items=50 `. The output will tell you next steps. It's also logged to file.
* To test the "nudger", run this for example: `wp newspack-content-migrator transform-blocks-nudge --num-items=30  --post-types=post,page`

---

- [x] confirmed that PHPCS has been run – yea, kinda!


src/Command/General/BlockTransformerCommand.php
```diff
@@ -68,59 +173,156 @@ public function cmd_blocks_decode( array $pos_args, array $assoc_args ): void {
 					'post_content' => $content,
 				]
 			);
-			WP_CLI::log( sprintf( 'Decoded blocks in %s', get_permalink( $post->ID ) ) );
+			$this->logger->log( $logfile, sprintf( 'Decoded blocks in ID %d %s', $post->ID, get_permalink( $post->ID ) ), Logger::SUCCESS );
```
eddiesshop, line 176: To prevent false positives, I think we should change this to `$updated = wp_update_post(...); if ( $updated) { $this->logger->log(...);` And only increment `$decoded_posts_counter` if `$updated === true`

src/Command/General/BlockTransformerCommand.php
```diff
@@ -68,59 +173,156 @@ public function cmd_blocks_decode( array $pos_args, array $assoc_args ): void {
...
 			}
+
 			wp_update_post(
 				[
 					'ID'           => $post->ID,
 					'post_content' => $content,
 				]
 			);
-			WP_CLI::log( sprintf( 'Encoded blocks in %s', get_permalink( $post->ID ) ) );
+			$this->logger->log( $logfile, sprintf( 'Encoded blocks in post ID %d  %s', $post->ID, get_permalink( $post->ID ) ), Logger::SUCCESS );
```
eddiesshop, line 242: Similar to comment from above, it'd be good to only display this information, if in fact we can confirm that `wp_update_post` was truly successful.

eddiesshop (changes requested):
I didn't actually test, just a straight ol' code review.

PR #488, opened 2024-04-18
Add support for image blocks

jorostoyanov:
## Overview

This PR is a continuation of https://github.com/Automattic/newspack-custom-content-migrator/pull/475.

It adds the removed support for image blocks parsing.

The logic is to loop through each image block of a Post and test if it can be used as a Featured Image. If such blocks don't exist or the images inside are not an actual Attachment, continue with the raw post content parsing.

---

### DIFF
--- a/src/Command/General/InlineFeaturedImageMigrator.php
+++ b/src/Command/General/InlineFeaturedImageMigrator.php
@@ -418,7 +418,7 @@ class InlineFeaturedImageMigrator implements InterfaceCommand {
 			// Get content for crawling.
 			$post_content = $wpdb->get_var( $wpdb->prepare( "SELECT post_content FROM $wpdb->posts WHERE ID = %d", $post_id ) );
 
-			$image_blocks = $this->get_image_blocks_from_post_content_blocks( parse_blocks( $post_content ) );
+			$image_blocks                       = $this->get_image_blocks_from_post_content_blocks( parse_blocks( $post_content ) );
 			$has_set_featured_image_from_blocks = false;
 
 			if ( ! empty( $image_blocks ) ) {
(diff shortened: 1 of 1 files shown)


iuravic (changes requested):
Here's a code readability suggestion. Writing a block of code and then adding comment _after_ the block makes it difficult to read and understand. I'd suggest one of these two improvements:
1. good -- move your comments to the top of your code, and instead of "Reaching here...", just use a comment like "// Check if post_content contains an image block with an 'id' attribute."
2. better -- put that block in a method which has such a nice name that it's practically self-documenting, e.g. public function does_content_have_img_block_with_id_attr( $post_content ): bool

Which ever you prefer personally, both are fine. I like 2. but no need to do it if you're more comfortable w/ 1.

But even more importantly, `$has_set_featured_image_from_blocks` has been set, but post has not actually been updated in that last if-block you wrote. Am I missing something here? Sorry if I am. But it looks like you detected that there _is_ an image block with attribute, but have not used that info anywhere after that.

iuravic:
Excellent, thanks @jorostoyanov, updating the comments will be great.

As for the line which sets the image, sorry I missed that part. The following comment purely a style point, so it's not a requirement for the PR, just a bit of discussion, I will leave it up to you to decide which of these two you'd like to use/keep.

Current:
```
if ( set_post_thumbnail( $post_id, $image_block['attrs']['id'] ) ) {
	$has_set_featured_image_from_blocks = true;
	break;
} else {
	WP_CLI::line( sprintf( '❌ Could not set Featured Image from Block. ID %d', $image_block['attrs']['id'] ) );
}
```

What I personally would propose, just from my own coding style perspective. This one looks more readable ti me, I can understand it in less time.
```
$has_set_featured_image_from_blocks = set_post_thumbnail( $post_id, $image_block['attrs']['id'] ) ? true : false;
if ( false === $has_set_featured_image_from_blocks ) {
	WP_CLI::line( sprintf( '❌ Could not set Featured Image from Block. ID %d', $image_block['attrs']['id'] ) );
}
```

But the this style choice is about you. What you think is nicer.

Once you've updated the comments, and if you decide to update this part, feel free to merge to master 👍 That's why I'm approving now, early.

iuravic (approved):

PR #502, opened 2024-05-22
Avoid returning an array with null entries

naxoc:
This fixes an edge case bug where the collations array was not "compacted" with array values.

---


kariae (approved):
Thank you!

PR #504, opened 2024-05-28
Medium Migrator: Change Inline Images Handler

jorostoyanov:
## Description

Medium uses `data-width` and `data-height` attributes for images in articles. These dimensions don't make sense in the context of uploading the images in WordPress. The Medium Migrator command was importing the images into the Media Library but was leaving the incorrect dimensions in the post content.

In addition, the caption related to the image was not updated on the uploaded Attachment.

To fix both issues the Migrator has been updated to:

 * parse the related image caption and use it as a caption for the Attachment
 * replace the whole image with a Gutenberg Image Block

## How to test

You will need a Medium Export Archive to run the migration locally. Ping me and I can send a sample to you 😊

---


naxoc (approved):
I didn't test it locally, but this looks good™️ !

It's a good improvement to get actual image blocks out of this :)

PR #506, opened 2024-05-31
Add/cli loader

naxoc:
This adds integration with the new separate repo for re-usable logic. 

For starters we use a super simple command to test on, but once we have this loading in this repo we can get more adventurous. The idea is that whenever we move over a class or function, we either deprecate it (that is ideal) or wrap it like I did in this PR (`\Newspack\MigrationTools\Command\AttachmentsMigrator::get_instance()->cmd_get_atts_by_years( $pos_args, $assoc_args );`

## How to test
This doesn't do much to existing code, so it's a bit hard to test, but something like this:
* Before switching to this branch:
  * Run `wp newspack-content-migrator attachments-get-ids-by-years` and note the output. (By far easiest on a pretty small site – I used the dummy data the Newspack plugin creates when the plugin is enabled).
* Switch to this branch
* Make sure to run `composer install`
* Now run the `wp newspack-content-migrator attachments-get-ids-by-years` command again. You should get the same output as before.
* Now run `wp newspack-migration-tools attachments-get-ids-by-years` and also ensure you get the same output.


kariae (approved):
LGTM, thanks!

PR #511, opened 2024-06-14
Delete config files that were never used

naxoc:
As far as I can tell we never ran on Travis or had JS that would need a package.json, so I deleted them.


kariae (approved):
Thanks!

PR #521, opened 2024-08-14
Use CAP logic class from migration tools repo instead

naxoc:
This pretty agressively deprecates the `NewspackCustomContentMigrator\Logic\CoAuthorPlus` class in favor of the [class now in the re-usable migration tools repo](https://github.com/Automattic/newspack-migration-tools/blob/trunk/src/Logic/CoAuthorsPlusHelper.php). The aggressive deprecation is on purpose so we don't work in two places. The class in the new repo is a 1:1 of the class I'm deprecating in this repo, so this PR is not about checking if the new class will work – but about it being included correctly. 

What I've done is just change the inclusion of the class to the new class. It's a lot of code to review, but it should be possible to do with a glance.

I've tested in a bunch of ways – even with the unit tests (they are broken without this PR, but this PR does not break them even more).

## How to test
Ideally you test this with something you are working on currently – so "actual" migration code that uses the class.  But here are some ideas if you don't have that kind of code laying around:

```bash
wp newspack-content-migrator co-authors-link-guest-author-to-existing-user --ga_id=<guest-user-id> --user_id=<user-id>
wp newspack-content-migrator co-authors-create-guest-author-and-add-to-post --post-id=$(wp post list --orderby=ID --order=desc --posts_per_page=1 --format=ids) --full_name='Gary BorkBorkzy'
```

---


eddiesshop (approved):
Everything you've done here makes sense. Good idea adding the `@deprecated` tags also!

PR #525, opened 2024-08-30
Move older pub-specific classes to an archive folder

naxoc:
This adds a README explaining what the folder does, and then moves all pub specific classes except for the ones included by `newspack-custom-content-migrator.php` to the archive. I have not otherwise touched the code.


iuravic (approved):

iuravic:
@naxoc , perhaps a good idea to bump up [the version number](https://github.com/Automattic/newspack-custom-content-migrator/blob/trunk/newspack-custom-content-migrator.php#L8). Since we're not really actively using it, I didn't make it a part of the review, but if you wanted, we can make this `1.6.0` in the plugin header comment.

PR #526, opened 2024-08-30
Add phpcs check on code in PRs

naxoc:
This GH action will run phpcs on the code touched in PRs. 

It has to do some juggling of json in the composer file to (lazily) avoid the private repositories in there, but since we only need the stuff in `require-dev` for running phpcs we should be fine.

## How to test
The easiest way to test is to push a change with some borked formatting in a PHP file and see the result, so there is a file that has that in here. I'll remove that before merging.


jorostoyanov:
LGTM ✅ 

Thanks 🙇

PR #530, opened 2024-09-03
Image block with custom link (from old PR#324)

iuravic:
Gutenberg helper class's `get_image` which returns an Image block, now gets the ability to not only use `$link_to_attachment_url` but also a `$custom_link`. The image block can normally use both these links.


jorostoyanov:
@iuravic Looks good to me! Thanks 🙇

jorostoyanov (approved):

PR #535, opened 2024-09-03
One way to stop instanciation of commands

naxoc:
I don't think there is much idea to instanciating _ALL_ command classes in the plugin setup. 

This is an idea on how we could switch the command add to be static and avoid all the instanciation of commmands.

I only put one example (`PostMigrator`) but we'd need to go through all commands and clean up.

## How to test

---


naxoc:
I'm closing this one in favor of: #536

PR #543, opened 2024-09-27
Make code php 8.3 compliant

naxoc:
## Description
We had a few lines that would fatal in PHP 8.3. This fixes that. It also changes the version phpcs checks against.

## How to test
- Run:
  -  `composer run-script phpcs -- -p --standard=PHPCompatibility --runtime-set testVersion 8.1 src`
  -  `composer run-script phpcs -- -p --standard=PHPCompatibility --runtime-set testVersion 8.1 tests`
  -  `composer run-script phpcs -- -p --standard=PHPCompatibility --runtime-set testVersion 8.3 src`
  -  `composer run-script phpcs -- -p --standard=PHPCompatibility --runtime-set testVersion 8.3 tests`
  - They should all come back with no errors.

### DIFF
--- a/phpcs.xml
+++ b/phpcs.xml
@@ -13,7 +13,7 @@
 	</rule>
 
 	<rule ref="PHPCompatibilityWP"/>
-	<config name="testVersion" value="7.4"/>
+	<config name="testVersion" value="8.3"/>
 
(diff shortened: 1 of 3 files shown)


jorostoyanov (approved):
Tested and works like a charm 🚀

PR #548, opened 2024-10-14
Remove GutenbergBlockManipulator, transformer and transformer command

naxoc:
## Description
See https://github.com/Automattic/newspack-migration-tools/pull/28


.github/workflows/phpcs.yml
```diff
@@ -17,7 +17,7 @@ jobs:
       - name: Check for PHP file changes
         id: check_php_changes
         run: |
-          CHANGED_PHP_FILES=$(git diff --name-only ${{ github.event.pull_request.base.sha }}...${{ github.event.pull_request.head.sha }} | grep '\.php$' || echo 'no files')
+          CHANGED_PHP_FILES=$(git diff --diff-filter=d --name-only ${{ github.event.pull_request.base.sha }}...${{ github.event.pull_request.head.sha }} | grep '\.php$' || echo 'no files')
```
eddiesshop, line 20: Was this intentional? Filter for only deleted files?

eddiesshop (commented):
Aside from the question on the phpcs.yml, is PostSelect a new class only in NMT? Meaning there was no equivalent in NCCM to remove?

.github/workflows/phpcs.yml
naxoc: When the `d` is lowercase it excludes instead of includes. Not the most intuitive, but what I'm trying to do is say "give me all files touched - but not the deleted ones because I can't check a file that doesn't exist".

naxoc:
>  is PostSelect a new class only in NMT? Meaning there was no equivalent in NCCM to remove?

Yes! I should have noted that. We have https://github.com/Automattic/newspack-custom-content-migrator/blob/trunk/src/Logic/Posts.php but I would love to start using some more clean functions.

PR #549, opened 2024-10-16
Remove unused files

naxoc:
Also clean up references to a file that was moved a while ago (the JsonIterator).

The mustache stuff was never used and the PHP.php file had no usages either.


publisher-specific-archive/LaSillaVaciaMigrator.php
```diff
@@ -119,6 +119,7 @@ class LaSillaVaciaMigrator implements InterfaceCommand {
 					'children' => array(),
 				),
 				array(
+					'name'     => 'Red Paisa',
```
jorostoyanov, line 122: There is a duplicate line here :)

src/Utils/PHP.php
jorostoyanov, line 1: I see this being used in `src/Command/General/S3UploadsMigrator.php`.

jorostoyanov (changes requested):
@naxoc Found one duplicate line and the PHP util class is still being used.

The rest looks good :)

src/Utils/PHP.php
naxoc: Great catch. I removed the usage

PR #565, opened 2024-12-04
Fix calling non-static Attachment logic class methods

iuravic:
## Description
Some non static Attachment logic class methods were called as if they were static, which threw fatals.


kariae:
LGTM! The method in question [is not static](https://github.com/Automattic/newspack-migration-tools/blob/5f2298c3259ee64615d5420c653e3c70c545bbf7/src/Logic/Attachments.php#L330), nice catch!

PR #575, opened 2025-01-21
Refactors shortcode functionality to NMT

iuravic:
## Main updates in this PR
- moves shortcode functionality from NCCM class `ContentFixerMigrator` to NMT (see [NMT PR 52](https://github.com/Automattic/newspack-migration-tools/pull/52))
- all history is kept together wit code owners

## Minor updates in this PR
- added a publisher specific migrator class


kariae:
I got confused for a second about the usage of `ShortcodeReplacementInterface`. I see the idea behind it now, but I think we don't need `SanDiegoVoiceAndViewpointMigrator` to implement `RegisterCommandInterface`.

My tiny concern is that we're making NMT rely on NCCM (with the `replace-callback` parameter) when it should be the other way around. Do you see my point?

kariae:
I agree with all the above. I didn't mean that we need migrators for specific shortcodes but that we could have a Logic method in NMT that we can call from a command here. Since we have commands on NMT too, I think both perspectives are okay!

I don't think we need a `replace-known-shortcodes-in-post-body` that's too granular, in my opinion.

kariae (approved):

PR #581, opened 2025-02-18
Optional post status param to InlineFeaturedImagesMigrator

iuravic:
## Description
Just a minor addition to the reusable migrator, added a --post-status optional param.
Also fixed some PHP sniffs.
Super simple logging private method used because existing logging method is deprecated and no longer working (breaking with error, actually).

---


iuravic:
Merging this one without review because it's just an addition of an optional parameter which is present in many of our existing migrators in exactly the same form, and this is needed for a publisher migration (see Asana OutSmart migration).

PR #589, opened 2025-04-08
fix: issue where EOF is reached prematurely, and `get_row` returns bool

eddiesshop:
## Description
There are some CSV files that have an extra line item at the end of it. It's empty though. But this means that `feof` returns `false` though. This then then leads to trying to get this row with `get_row` and the `fgetcsv` call returning `false` instead of an expected array.

These changes will now let `get_row` to return `bool` (which can happen when calling `fgetcsv`), and when a `bool` is received, it means we have reached the end of file, there are no more rows to process, so we should either return/exit out of the current function or throw an Exception.

## How to test
1. Open the attached CSV:
> $csv = new \NewspackCustomContentMigrator\Utils\CommonDataFileIterator\CSVFile( $path_to_file );
2. Get an iterator:
> $it = $csv->getIterator();
3. Loop through, see that an PHP fatal is thrown before reaching the end of file
> foreach( $it as $row ) {
>    var_dump( $row );
> }
4. Check out this PR/branch
5. Repeat steps 1-3, notice that the end of file is reached and no fatal is thrown.
[test.csv](https://github.com/user-attachments/files/19656098/test.csv)

---

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -55,12 +55,12 @@
             "source": {
                 "type": "git",
                 "url": "https://github.com/Automattic/newspack-migration-tools.git",
-                "reference": "6c7189e93a4cdcfc8022354766293d69241dd926"
+                "reference": "921a2c9defe755641ebdce3753dd76ea61c38d5b"
             },
             "dist": {
                 "type": "zip",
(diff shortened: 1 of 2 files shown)


jorostoyanov (approved):
Had the same issue the other day 😅 LGTM 🙌

PR #605, opened 2025-07-24
America Mag: Legacy Podcasts Migrator

jorostoyanov:
## Description

This PR adds a new command to add legacy podcasts as an embed block in the related post content.

## Synopsis

```
wp newspack-content-migrator america-mag-migrate-legacy-podcasts-links
    --start=<index to start from>
    --end=<index to end with>
```

Upon completion, all Posts should have a new `core/embed` blocks as the first block in the content with the appropriate Attachment URL.

A CSV is generated that contains a list of all updated posts.


jorostoyanov:
> @jorostoyanov - very cool, but as of now we're not doing NCC block converter... Will this work with just plain html in post_content? Something as simple as just pre-pending the url with `\n` at the top of post_content is fine:
> 
> ```
> http://...link to mp3
> 
> the post content ....
> ```

It will, but it's not an issue to have both block and classic content in the post_content.

jorostoyanov:
Hey @ronchambers 👋 

Your statement about NCC is absolutely correct! There is an idea to change this behaviour, so it doesn't only check if the content starts with block HTML comment, but that's another topic.

In order to go around that, we can just add a blank line in the beginning of the post content, just before the Embed block. This way, we will have native Embed block and still allow to later convert the whole Post Content with NCC ([Reference](https://github.com/Automattic/newspack-custom-content-migrator/pull/605/commits/5b71f16842eea26e10c18b95d045aa1a6ef8c2a6))

Let me know if this looks good, and I will merge it :)

